PR #2, opened 2021-03-03
add input style

schizobulia:


Kaiden42:
Hello, thanks for your PR. Could you add a description of what you want to achieve with this PR? What is the purpose of having another style for the text_input?

schizobulia:
Modify the color of border_colord when text_input gets focus，such as [bootstrap](https://getbootstrap.com/docs/5.0/forms/overview/#overview)

Kaiden42:
Basically, styling is a matter of the user.

The "bootstrap colors" in this crate are actually just a kind of quickstart feature to give e.g. badges or cards a meaning through the color coding during the early development phase of the application without having to implement these styles yourself.

I don't know if we should also include predefined styles for more widgets in this crate to experiment with. Since this is now a community crate, I'll leave this discussion open for the community 🙂 .

13r0ck:
It is an interesting idea for sure, though it might be more useful for the [Iced docs](https://github.com/iced-rs/iced-rs-docs) to contain a part with a few guides for different example styles in Iced.

genusistimelord:
I do not believe this should be in this crate. However, you could make a New crate that deals with predetermined styling for all widgets in Iced, Iced_aw and Iced_audio. As I think a Separate library would be better for this as it needs to cover way more than Iced_aw. Maybe Called it Iced_Styles. If you would like to make such a crate, I am willing to help you get started and I can even help get it added to the Iced Repository once it’s Done. @schizobulia

PR #4, opened 2021-03-05
[WIP] Menu widget

Kaiden42:
First draft of a `Menu` widget.

menu

Open TODOs:
* [x] Disable entries. This could be achieved by changing `Message` to `Optional<Message>`.
* [x] Toggle-able entries. This could be achieved by adding a `Entry::Toggle` or something to `Entry`.
* [x] Cool looking separators between the entries.
* [x] Bouncing on the edge of the window or else the menu is cut off.
* [ ] ~Add some internal scrollable if the list of entries is to high for the window.~ This seems to be realy hacky to implement. Nevertheless it would require for the `scrollable::State` to be hashable and maybe to fix [this](https://github.com/hecrj/iced/issues/765) first.
* [ ] Maybe a cooler default style. Some suggestions?
* [ ] Add some way of keyboard control. In common, the `Alt` + `[First character of section label]` combination is used to pop up the content of the specific section. This would require to change the `Text` label of the `Section` to a normal `String` (or `&str`)  to be able to know the first character of the label.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ floating_button = ["button"]
 glow = [] # TODO
 icon_text = []
 icons = []
+menu = ["icon_text"]
 modal = []
 tab_bar = []
 tabs = ["tab_bar"]
@@ -34,6 +35,7 @@ default = [
     "date_picker",
     "color_picker",
     "floating_button",
+    "menu",
     "modal",
     "tab_bar",
     "tabs",
@@ -65,6 +67,7 @@ members = [
     "examples/color_picker",
     "examples/date_picker",
     "examples/floating_button",
+    "examples/menu",
     "examples/modal",
     #"examples/tab_bar",
     #"examples/tabs",
--- a/examples/menu/Cargo.toml
+++ b/examples/menu/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "menu"
+version = "0.1.0"
+authors = ["Kaiden42 <gitlab@tinysn.com>"]
+edition = "2018"
+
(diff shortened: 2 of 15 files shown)


yusdacra:
This is very nice! I was thinking of making something like this for my chat app since using a PickList as a menu was hacky, but looks like you were faster :P

It seems we can use anything that can `Into<Element<_>>` for menu entries which is appreciated. Something that would be nice is methods on `Menu` that would help with prototyping, eg. perhaps a method that takes an array of strings for easier creation of text-only `Menu`s etc.

genusistimelord:
Hmm for styling lets just stick with a workable default. Also scroll-able stuff should be fixed to render properly since a few versions back. a few keyboard controls would be nice probably can try to help implement them later. Also the pure edition might also fix the state hashing requirements for the scrollbars.

Kaiden42:
@genusistimelord For now, I would prefer to wait for the support of the real system menus ([iced#1047](https://github.com/iced-rs/iced/pull/1047), [iced#114](https://github.com/iced-rs/iced/issues/114)). What do you say?

genusistimelord:
> @genusistimelord For now, I would prefer to wait for the support of the real system menus ([iced#1047](https://github.com/iced-rs/iced/pull/1047), [iced#114](https://github.com/iced-rs/iced/issues/114)). What do you say?

yeah thats fine. was more or less stating how the current changes in iced have already changed some of the requirements.

genusistimelord:
I am closing this since menu now exists. Decided to allow it since this cant be supported fully in other ways. we might not see its support for a long time if any.

PR #6, opened 2021-03-06
feat(modal): state is_shown method

yusdacra:
Sorry for the small PR, but it seems I forgot about this while doing the other PR (i only really needed this now).

This adds a `is_shown` method to the `modal::State` to query if the modal will be shown or not.


Kaiden42:
Don't worry about that. Looks good, thanks! :+1:

PR #9, opened 2021-03-24
Overwrite `overlay` method in Widget implementation for Tabs

thenlevy:
The implementation of `Widget` for `Tabs` needs to overwrite the `overlay` method, otherwise picklist cannot be used inside of tabs.

### DIFF
--- a/src/native/tabs.rs
+++ b/src/native/tabs.rs
@@ -417,6 +417,20 @@ where
         self.width.hash(state);
         self.height.hash(state);
     }
+
+    fn overlay(
+        &mut self,
+        layout: Layout<'_>,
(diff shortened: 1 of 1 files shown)


Kaiden42:
Thanks for your PR. :+1: 
You used `unwrap` in a couple of places. Even though `unwrap` provides fast coding, I prefer using `expect` for code in production to get better error output.

Could you replace the places with `unwrap` with `expect` and an appropriate error message? Otherwise I can also quickly take care of it.

thenlevy:
> Could you replace the places with `unwrap` with `expect` and an appropriate error message? Otherwise I can also quickly take care of it.

Done :+1:

Kaiden42:
Thank you very much :slightly_smiling_face:

PR #10, opened 2021-04-06
Use latest released version of `iced` on crates.io

hecrj:
See https://github.com/hecrj/iced/pull/806.


Kaiden42:
Awesome! Thanks! :+1: 

This allows iced_aw to be made available via Crates.io.

PR #11, opened 2021-04-17
Add ``NumberInput`` widget

hangleang:
Hi there. This is the implementation of ``NumberInput`` widget. Sorry for committing lately, I was very busy on my work recently. 

Feel free to give me some idea to improve this widget. Thanks you guys....

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ modal = []
 tab_bar = []
 tabs = ["tab_bar"]
 time_picker = ["chrono", "icon_text", "iced_graphics/canvas"]
+number_input = []
 
 default = [
     "badge",
@@ -42,6 +43,7 @@ default = [
 
 [dependencies]
 iced_style = "0.3"
+num-traits = "0.2.14"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 iced_native = "0.4"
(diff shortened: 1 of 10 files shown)


Kaiden42:
Thanks for your PR! :slightly_smiling_face: 
> Sorry for committing lately, I was very busy on my work recently.

Don't worry, I was busy as well.

Looks pretty good so far. However, I have two small suggestions for improvement:
I saw that the new `number_input` widget depends on the crate `num-traits`. Could you, similar to the crate `chrono`, mark this as optional? For the `number_input` feature you can then say that it depends on the `num-traits` "feature". Similar to the `date_picker` feature. This way the dependency will only be compiled if the `number_input` feature is enabled. You can read more about this [here](https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies).

Additionally, have you thought about floating point support? These are currently only partially supported. I was able to change the type of the `number_input` example from `u8` to `f32`, but it caused strange effects. I am not able to enter a decimal point into the field myself, but I am able to generate floating point numbers via the increment/decrement by setting the step size to `0.5`. If I then delete the decimal point, I even get beyond the maximum value.
Example: I set the value to `26` and increment by `0.5`. This gives me `26.5`. If I now remove the decimal point I am at `265`, whereas only a maximum of `255.0` is allowed.

hangleang:
Done 😀

Kaiden42:
Awesome! Thank you for your work. :slightly_smiling_face: 

I've now merged your PR. However, I noticed two minor issues, but they can be addressed in the future.
- With decimal support I'm not able to type the decimalpoint at the end of the number. This could be caused by the cut-off of the ".0" decimal.
- Support for web would be great in the future.

PR #14, opened 2021-05-03
Fix README's cargo deps

### LINKED ISSUE #13: Cannot compile TabBar example
I want to build an app starting with the TabBar as a base to work off. Following the README, my cargo dependencies are:
```toml
[dependencies]
iced =  { git = "https://github.com/hecrj/iced", rev = "12c0c18d662d2b817b559b94c71d18e122c76990" }
iced_aw = { git = "https://github.com/iced-rs/iced_aw", branch = "main", default-features = false, features = ["tab_bar"] }
```

Note that I'm using `iced-rs/iced_aw` instead of `kaiden42/iced_aw` and specified the `main` branch and the `tab_bar` feature.

I then copy-pasted the source of [the TabBar example](https://github.com/iced-rs/iced_aw/blob/main/examples/tab_bar/src/main.rs) into my main.rs, but get the following errors:
```
error[E0308]: mismatched types
   --> src\main.rs:127:32
    |
127 |                     .tab_width(Length::Shrink)
    |                                ^^^^^^^^^^^^^^ expected enum `iced_core::length::Length`, found enum `Length`
    |
    = note: perhaps two different versions of crate `iced_core` are being used?

error[E0277]: the trait bound `iced_native::element::Element<'_, Message, iced_graphics::renderer::Renderer<iced_wgpu::backend::Backend>>: From<iced_aw::native::TabBar<Message, iced_graphics::renderer::Renderer<_>>>` is not satisfied
    |
118 |             .push(
    |              ^^^^ the trait `From<iced_aw::native::TabBar<Message, iced_graphics::renderer::Renderer<_>>>` is not implemented for `iced_native::element::Element<'_, Message, iced_graphics::renderer::Renderer<iced_wgpu::backend::Backend>>
````

I tried to solve the first by removing the `rev` from the cargo dependency, but no luck. 
The second error is totally beyond my Rust competency.

mon:
- Iced uses 0.3, not a git commit
- Repo is now under the `iced-rs` org
- Cargo needs to be told the repo lives in `main`, else it will try `master` and fail

Will close #13

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,8 @@ Every widget is hidden by a feature gate. This allows you to cherry pick the wid
 Include `iced_aw` as a dependency in your `Cargo.toml`:
 ```toml
 [dependencies]
-iced =  { git = "https://github.com/hecrj/iced", rev = "12c0c18d662d2b817b559b94c71d18e122c76990" }
-iced_aw = { git = "https://github.com/kaiden42/iced_aw", default-features = false, features = [...] }
+iced = "0.3"
+iced_aw = { git = "https://github.com/iced-rs/iced_aw", branch = "main", default-features = false, features = [...] }
 ```
 
 **Why not Crates.io?**


Kaiden42:
Thanks :+1:

PR #16, opened 2021-05-10
Add a grid widget

Luni-4:
Hello, everyone! This is my first contribution here, thanks a lot for the project! :smiley: 

I decided to open this PR in this repository after having read this comment https://github.com/hecrj/iced/pull/762#pullrequestreview-605699978 and it is a follow-up of https://github.com/hecrj/iced/pull/189 

The widget was entirely implemented by @daxpedda, I just moved some code around and implemented the example. Unfortunately I wasn't able to create the video that explains how the example works at runtime because I don't know how to do that. May someone please do that for me? Thanks! 

I haven't added any info to the README either because, even in this case, I don't know the template that should be used to explain a new widget.

I hope this will be appreciated and thanks in advance for your review!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,7 @@ members = [
     "examples/color_picker",
     "examples/date_picker",
     "examples/floating_button",
+    "examples/grid",
     "examples/modal",
     #"examples/tab_bar",
     #"examples/tabs",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ colors = []
 date_picker = ["chrono", "lazy_static", "icon_text"]
 color_picker = ["icon_text", "iced_graphics/canvas"]
 floating_button = ["button"]
+grid = []
 glow = [] # TODO
 icon_text = []
 icons = []
@@ -35,6 +36,7 @@ default = [
     "date_picker",
     "color_picker",
     "floating_button",
+    "grid",
     "modal",
     "tab_bar",
     "tabs",
--- a/examples/grid/Cargo.toml
+++ b/examples/grid/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "grid"
+version = "0.1.0"
+authors = ["daxpedda <daxpedda@gmail.com>", "Luni-4 <luni-4@hotmail.it>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+iced =  "0.3"
+iced_aw = { path = "../..", default-features = false, features = ["grid"] }
--- a/examples/grid/src/main.rs
+++ b/examples/grid/src/main.rs
@@ -1,0 +1,88 @@
+use iced::{
+    button, scrollable, Align, Button, Color, Column, Container, Element, Length, Sandbox,
+    Scrollable, Settings, Text,
+};
+
+use iced_aw::Grid;
+
+// Number of columns for the grid
+const COLUMNS: usize = 2;
+
+fn main() -> iced::Result {
+    GridExample::run(Settings::default())
+}
+
+#[derive(Debug, Clone)]
+enum Message {
+    AddElement,
+}
+
+struct GridExample {
+    element_index: usize,
+    button_state: button::State,
+    scrollable_state: scrollable::State,
+}
+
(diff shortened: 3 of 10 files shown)


Kaiden42:
Thank to both of you for this cool widget! :slightly_smiling_face: 

There is a small issue with your example code in the documentation. You need to annotate the message type like [here](https://github.com/iced-rs/iced_aw/blob/7a6918a50fa6784ce7d11348b71ea4331d9d6200/src/native/badge.rs#L24). You can run doc-test simply with `cargo test`. But I'm wondering why the GitHub workflow wasn't running to check this. :thinking: 
```Rust
---- src/native/grid.rs - native::grid::Grid (line 16) stdout ----
error[E0282]: type annotations needed for `iced_aw::native::Grid<'_, Message, iced_native::renderer::Null>`
 --> src/native/grid.rs:20:12
  |
7 | let grid = Grid::with_columns(2)
  |     ----   ^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `Message`
  |     |
  |     consider giving `grid` the explicit type `iced_aw::native::Grid<'_, Message, iced_native::renderer::Null>`, where the type parameter `Message` is specified

error: aborting due to previous error
```

> Unfortunately I wasn't able to create the video

Don't worry about that, I can create this video by myself. If you would like to be able to do it for further contributions, you need a screen recording program for this. There are a lot of such programs like, for example, [OBS](https://obsproject.com) which would be an overkill for this small tasks, or a more lightweight program like [Kazam](https://wiki.ubuntuusers.de/Kazam/) (Linux only) which I am using. Since GitHub can't handle video files, you need to convert the video to a gif. You could use [GifCurry](https://github.com/lettier/gifcurry) (Linux / Mac only) for this.

> I haven't added any info to the README either because, even in this case, I don't know the template that should be used to explain a new widget.

Haha, me neither. :smile: The README should need a rework since it is growing a lot (with a lot of images). I will rework it in the upcomming days if I can find the time to do it.

Luni-4:
> Thank to both of you for this cool widget! slightly_smiling_face
> 
> There is a small issue with your example code in the documentation. You need to annotate the message type like [here](https://github.com/iced-rs/iced_aw/blob/7a6918a50fa6784ce7d11348b71ea4331d9d6200/src/native/badge.rs#L24). You can run doc-test simply with `cargo test`. But I'm wondering why the GitHub workflow wasn't running to check this. thinking
> 
> ```rust
> ---- src/native/grid.rs - native::grid::Grid (line 16) stdout ----
> error[E0282]: type annotations needed for `iced_aw::native::Grid<'_, Message, iced_native::renderer::Null>`
>  --> src/native/grid.rs:20:12
>   |
> 7 | let grid = Grid::with_columns(2)
>   |     ----   ^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `Message`
>   |     |
>   |     consider giving `grid` the explicit type `iced_aw::native::Grid<'_, Message, iced_native::renderer::Null>`, where the type parameter `Message` is specified
> 
> error: aborting due to previous error
> ```

Thanks for the hint! It should be fixed now,  running `cargo test` locally doesn't show the problem anymore. About the failing `GitHub Actions`, I guess it's my fault because the actions is currently disabled for my account, so it's not running on my PRs for now.


> Don't worry about that, I can create this video by myself. If you would like to be able to do it for further contributions, you need a screen recording program for this. There are a lot of such programs like, for example, [OBS](https://obsproject.com) which would be an overkill for this small tasks, or a more lightweight program like [Kazam](https://wiki.ubuntuusers.de/Kazam/) (Linux only) which I am using. Since GitHub can't handle video files, you need to convert the video to a gif. You could use [GifCurry](https://github.com/lettier/gifcurry) (Linux / Mac only) for this.

Great, I will try to follow your suggestions for further contributions then, thanks a lot! :smile: 

> Haha, me neither. smile The README should need a rework since it is growing a lot (with a lot of images). I will rework it in the upcomming days if I can find the time to do it.

I see, no problem then, take your time! :+1:

Kaiden42:
>  I guess it's my fault because the actions is currently disabled for my account, so it's not running on my PRs for now.

Yes, this could be the explanation for this. :slightly_smiling_face: 

> Great, I will try to follow your suggestions for further contributions then, thanks a lot! smile

I'm glad I was able to help.

Thank you really much for your work. :+1:

PR #26, opened 2021-06-21
Submitting Fixes for Number_input #18 and Color Picker issue not yet reported.

### LINKED ISSUE #18: NumberInput panics when overwriting selected number
The number input panics when selected text is overwritten.
```
thread 'main' panicked at 'slice index starts at 1 but ends at 0', /home/user/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/vec/mod.rs:1735:36

```

Steps to reproduce this issue:
1. Create a NumberInput that starts with a value like `5` for example.
2. Select `5` in the TextInput part of the NumberInput.
3. Type another number to overwrite the selection.
4. Panic

Maybe this is caused by the text being empty.

genusistimelord:
I have updated iced_aw and fixed #18  and also Fixed the Color pickers Cancel button issue where Cancel would throw Submit.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ tabs = ["tab_bar"]
 time_picker = ["chrono", "icon_text", "iced_graphics/canvas"]
 wrap = []
 number_input = ["num-traits"]
-
+selection_list = []
 
 default = [
     "badge",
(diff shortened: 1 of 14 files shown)


genusistimelord:
Just Fixing some spelling mistakes I made.

Kaiden42 (approved):
Thank you very much! I guess I always overlooked the error with the ColorPicker.

Issue #35, opened 2021-07-02
Selection List Make Positioning and Width/Height issues.


genusistimelord:
I am currently Struggling to figure out how to fix both the Height/width issue as well as keep the positioning Constant. In this way id prefer it was not an overlay but i think there is not many ways to do this otherwise. So i am looking to see if you can give me some help @Kaiden42. Been scratching my head since this morning over this issue.

So if you place something before it like a empty Space it will lay over that space OK. IF the space is to the Top most of the container and the lists height - position then it will render above the slider box. Still trying to figure out how the Location/sizing is handled in Iced......

Kaiden42:
The question is: should this widget be implemented as an overlay? If not, it could just be implemented as a normal widget.

genusistimelord:
It probably should be a normal widget and not a overlay.

genusistimelord:
So I managed to Somehow convert it to a Regular Widget XD. However I'm still having a few issues ATM with it. like i cant get the layout to return the correct sizes. or the rendering portion to give the correct layout bounds. I am getting there.

Kaiden42:
> So I managed to Somehow convert it to a Regular Widget XD

That's great! :+1: 

> However I'm still having a few issues ATM with it. like i cant get the layout to return the correct sizes. or the rendering portion to give the correct layout bounds.

Do I understand correctly that you were able to fix the problem now with #37?

genusistimelord:
in my latest commits It is now fixed. Just need to fix the other small things you pointed out.

PR #37, opened 2021-07-06
Changed Selection List from a Overlay to a Widget. Corrected Layout…

### LINKED ISSUE #31: Selection list in split (invisible/non-displaying)
Hello, I have trouble with the selection list in a split.

https://github.com/barvirm/Radventurer

```
// self.paths = vec!["FIRST ROW IN LIST".to_string(), "SECOND ROW IN LIST".to_string()]
let first = Container::new(Text::new("First"))
    .width(Length::Fill)
    .height(Length::Fill);

let selection_list = SelectionList::new(
    &mut self.selection_state,
    &self.paths[..],
    Some(self.selected_path.clone()),
    Message::OnFileClick,
);
    
let split = Split::new(&mut self.split_pane, first, selection_list, Message::OnResize);

```

image

genusistimelord:
…Positioning and Size issues.

This should also Fix #31 and #35


genusistimelord:
image
@
This should fix the layout issues with the test program RAdventurer from barvirm

src/native/overlay/list_menu.rs
```diff
@@ -1,417 +0,0 @@
...
-impl<'a, T, Message, Renderer> From<List<'a, T, Renderer>> for Element<'a, Message, Renderer>
-where
-    T: ToString + Clone,
-    Message: 'a,
-    Renderer: 'a + self::Renderer,
-{
-    fn from(list: List<'a, T, Renderer>) -> Element<'a, Message, Renderer> {
-        Element::new(list)
-    }
-}
```
Kaiden42, line 417: This file could now be placed in a submodule of the selection list instead of the overlay module

src/style/selection_list.rs
```diff
@@ -1,34 +1,62 @@
...
+    /// The List Label Text Select Background Color
+    pub selected_background: Background,
+    /// The Containers Width
+    pub width: Length,
+    /// The Containers height
+    pub height: Length,
+    /// The padding Width
+    pub padding: u16,
+    /// The Text Size
+    pub text_size: u16,
```
Kaiden42, line 28: This is a little unusual. Normally, the style of the widgets do not change the behavior of the layout. You could place this fields in the struct of the widget and implement setters to modify this values.

Take a look at the [Column](https://github.com/hecrj/iced/blob/master/native/src/widget/column.rs) widget as an example.

src/native/overlay/list.rs
```diff
@@ -0,0 +1,168 @@
...
+    for Element<'a, Message, Renderer>
+where
+    T: ToString + Clone,
+    Message: 'a,
+    Renderer: 'a + self::Renderer,
+{
+    fn from(list: List<'a, T, Message, Renderer>) -> Element<'a, Message, Renderer> {
+        Element::new(list)
+    }
+}
```
Kaiden42, line 168: This file could now be placed in a submodule of the selection list instead of the overlay module, as it is no overlay anymore.

Kaiden42 (changes requested):
Looks good. However, I had some small requests.

src/style/selection_list.rs
genusistimelord: the issue was the only way I could make it work without requiring List to be a overlay was to do 

https://github.com/iced-rs/iced_aw/blob/c69a4a906fdfb9892403ca37edbcade59b38aea4/src/native/selection_list.rs#L75

which means I either need to pass the width and height and such in Directly OR pass it indirectly. In which this case the Style stuff was designed to be similar to that of CSS or at least I Thought. This seemed to be the better way of handling it since i do not get direct access into the containers children after they are created. So i cant actually Set these for the list except at the very start of the new function.


If you prefer it being in the new function Or see another way to handle it please let me know and I will fix the issue.

src/native/overlay/list.rs
genusistimelord: I will move it to its own module

src/style/selection_list.rs
Kaiden42: Yes, this is indeed a little problem. Since I don't have a good solution for it at the moment either, we'll keep it for now and deal with it later.

Kaiden42 (approved):

PR #43, opened 2021-09-17
Build rustdoc to github pages

Luro02:
This crate is not available on crates.io, so you can not read its docs on docs.rs.

I have a similar crate that can not be published to crates.io, so I just use GitHub pages for hosting the docs and linking to it from the readme:

https://github.com/Luro02/array-map/tree/gh-pages

and this is the link to the docs: https://luro02.github.io/array-map/

I think this would be nice to have for this crate, so I created this PR.

---

I don't know if you want to use a real domain name, which you can do by going to the settings of the repository under GitHub pages there is a field to add a custom domain and then adding this to the yaml:

```diff,yml
          publish_branch: gh-pages
          publish_dir: ./target/doc
+         cname: iced_aw.iced.rs
```

Otherwise, one could access the docs through this url: https://iced-rs.github.io/iced_aw/

### DIFF
--- a/.github/workflows/rustdoc.yml
+++ b/.github/workflows/rustdoc.yml
@@ -1,0 +1,37 @@
+name: Deploy Docs to GitHub Pages
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  release:
+    name: GitHub Pages
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v1
+
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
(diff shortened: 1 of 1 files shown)


Luro02:
(not sure why CI is broken, since I did not change any code)

genusistimelord:
@Kaiden42 any reason why we should not have automated document building until we release on crates.io?

Kaiden42:
@genusistimelord No, it's actually an excellent idea. This would also give us documentation of unstable versions if someone wants to include iced_aw directly from GitHub.

Kaiden42 (approved):
Thank you very much for your PR! It is a great improvement to the usability of `iced_aw` as a developer.  :slightly_smiling_face: 

I apologize for not getting around to looking at it until so late.

PR #46, opened 2022-02-08
Update Beta to latest iced

Kaiden42:
Fixing following breaking changes:

- `iced_web` was moved into it's own repository
- A widget now doesn't have it's own renderer trait
- Rendering has changed
- Pulling mouse interaction out of renderer implementation into it's own method


mettz:
Hey, @Kaiden42, I would like to use the tab_bar and tabs widgets on a project of mine and I saw this pr so I thought that I could directly help with the fixes. I checked out your pr and hopefully fixed the breaking changes within `tab_bar` and `tabs`. I think I could directly push my changes here if you give me access or I could open a new pr and you might be able to merge them somehow. I'm not an expert here so I'll do what you think is better

PR #47, opened 2022-02-11
Removed no longer needed <S> from style functions

genusistimelord:
This is to fix the style functions in a few widgets. The error occurs as there is no longer a use for `<S>` so `<S>` is no longer preset by the compiler hence you must manually set it even though it does nothing now.


Kaiden42 (approved):
Thanks! :+1: 

It's interesting that neither rustc nor clippy were complaining about the unused generic :thinking:

PR #48, opened 2022-02-13
Show which item is selected in `SelectionList`

emann:
This PR implements an easy way to tell which item in a `SelectionList` is currently selected by tracking the index of the currently selected item and styling it accordingly.

Previously, `Style.selected_text_color` and `Style.selected_background` defined the styling of the item that the mouse was hovering over. These have been renamed to `Style.hovered_text_color` and `Style.hovered_background`, respectively, and `Style.selected_text_color` and `Style.selected_background` now accurately refer to the styling applied to the currently **selected** item.

I have verified that the `selection_list` example still works - no code changes were required.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ default = [
 ]
 
 [dependencies]
-iced_style = { git = "https://github.com/hecrj/iced.git", rev = "adce9e0" }
+iced_style = { git = "https://github.com/iced-rs/iced.git", rev = "adce9e0" }
 num-traits = { version = "0.2.14", optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-iced_native = { git = "https://github.com/hecrj/iced.git", rev = "adce9e0" }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -50,12 +50,12 @@ default = [
 ]
 
 [dependencies]
-iced_style = { git = "https://github.com/hecrj/iced.git", rev = "adce9e0" }
+iced_style = { git = "https://github.com/iced-rs/iced.git", rev = "adce9e0" }
```
Kaiden42, line 53: This change breaks every example (at least for me). This change would need to be taken on every example as well 
Would you like to open up another PR with this changes and remove this change from this PR for now?

genusistimelord: yeah id request that he removes this change from his PR so we can go fix these in a separate PR.

Kaiden42 (changes requested):
Thank you very much for your PR! :slightly_smiling_face: 

However, there are two small change requests that need to be solved before merging.

PR #52, opened 2022-05-03
Beta

genusistimelord:
- Updated to Iced `0.4`.
- Changed Cargo.toml to use a dependency method similar to WGPU which allows us to easily modify Version/git and rev without removing them completely. Good for Beta testing changes against Iced.
- Fixed Cargo Warnings.


Kaiden42 (approved):
Thank you very much! :+1:

PR #67, opened 2022-11-14
feat: make the floating button round

Sherlock-Holo:
by apply the custom button style with 200.0 radius, the button now looks like a round and similar to the example gif

the gif round is 
image
and the round created by PR
image


genusistimelord (approved):
looks good thanks

PR #72, opened 2022-11-28
Fix rounded header bottom of card. It looks cleaner square.

flomang:
This is purely subjective, but the header for a card looks cleaner with a square bottom. Or at least this could be used to allow that bottom border to be configurable as well.


flomang:
Done.

genusistimelord (approved):
thank you.

PR #74, opened 2022-12-12
Compile against iced_rs 0.6.0

yeastplume:
Note I have not tested this extensively, only branched, updated and ensured it compiles against my own project at https://github.com/mimblewimble/grin-gui

* Update version for iced, iced_native and iced_graphics
* Deal with addition of `BorderRadius`
* Deal with change to  `overlay` trait function of `Widget`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "iced_aw"
-version = "0.2.0"
+version = "0.3.0"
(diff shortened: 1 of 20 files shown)


yeastplume:
Note there are also some `chrono` deprecation warnings, but resolving them requires decisions that I'll leave to the maintainers.

genusistimelord (approved):
Looks Good. I am not worried about the chrono warnings atm. Thank you for your contribution.

PR #78, opened 2023-01-15
Upgrade to iced 0.7

carl-anders:
Only minor changes are needed to compile against `iced` 0.7.
I have tested all the programs in examples/ and they all compile, run and seem to work.
Updated `iced_native` to 0.8, `iced_graphics` to 0.6, `iced_style` to 0.6, `iced` to 0.7.

Examples `time_picker` and `date_picker` give deprecation warnings for `chrono`, but that is unrelated to the changes made.
There's some `cargo fmt` changes that should probably be done too, but I only applied formatting to code I changed in this PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,17 +60,17 @@ lazy_static = { version = "1.4.0", optional = true }
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies.iced_native]
 #git = "https://github.com/iced-rs/iced.git"
 #rev = "8221794"
-version = "0.7"
+version = "0.8"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies.iced_graphics]
 #git = "https://github.com/iced-rs/iced.git"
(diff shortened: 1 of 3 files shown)


genusistimelord (approved):
looks good thank you.

carl-anders:
@genusistimelord If I do further changed to iced_aw, should I `cargo fmt` the files I changed or try to keep the old formatting where possible?

genusistimelord:
yes we would appreciate that. cargo fmt is what we generally use.

GyulyVGC:
That's amazing, thanks @carl-anders.

Now that we are again in parallel with iced latest version, what should we do to use iced_aw?
I guess that the answer is to keep using the git feature in the cargo.toml; however, are you guys planning some workaround to getting it published on crates.io even if the package owner is gone? If there is no way to ask him to give someone else the privileges to publish a version, should we make an alternative repo and publish that to crates.io? 

What do you think? @genusistimelord

PR #79, opened 2023-01-16
Fix doctests by using the native type version

carl-anders:
Doctests seems to have been broken for at least a few commits. I noticed it before I started the 0.7 PR but decided not to change it in the 0.7 PR, as it did not bring any new errors.

The problem is that `Theme` has been added, and I couldn't figure out how to make `Null` take a theme easily. It was a much smaller change to simply use the native type in the doctests for the native elements. It shouldn't matter which type we import as the imports are hidden in the documentation anyway and are just used to make the doctests run successfully. It also makes a ton of sense to use the native type in the native files.

With this PR, `cargo test` will be successful again.

I also ran `cargo fmt` on the files I changed. But there are still other files which have not been changed in this PR which will fail the `cargo fmt` test.

### DIFF
--- a/src/native/badge.rs
+++ b/src/native/badge.rs
@@ -17,10 +17,10 @@ const BORDER_RADIUS_RATIO: f32 = 34.0 / 15.0;
 ///
 /// # Example
 /// ```
-/// # use iced_aw::style::badge;
 /// # use iced_native::{widget::Text, renderer::Null};
+/// # use iced_aw::native::badge;
 /// #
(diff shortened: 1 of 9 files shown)


genusistimelord:
Ahh yeah the issue we had before is Null on Iced did not support the stuff needed. So I'm guessing Null was finally updated. Thank you for this!

genusistimelord (approved):
Everything Code wise looks good.

carl-anders:
The cargo fmt errors are not for my changes but for code in other files.

genusistimelord:
yeah its fine, they were just confusing as to why they exist lol.

PR #80, opened 2023-01-16
Fix chrono deprecation warnings by replacing with equivalent non-deprecated code

carl-anders:
Many of the deprecation warnings are of functions that are very simple internally, and can be replaced by using the same code called in the now deprecated function.

* [`NaiveDate::from_ymd(_)`](https://github.com/chronotope/chrono/blob/9e5eb49838835337f136472fc5103357113b64aa/src/naive/date.rs#L255) is equivalent to `NaiveDate::from_ymd_opt(_).expect("invalid or out-of-range date")`
* [`NaiveTime::from_hms(_)`](https://github.com/chronotope/chrono/blob/9e5eb49838835337f136472fc5103357113b64aa/src/naive/time/mod.rs#L217) is equivalent to `NaiveTime::from_hms_opt(_).expect("invalid time")`
* [`Local::today().naive_local()`](https://github.com/chronotope/chrono/blob/9e5eb49838835337f136472fc5103357113b64aa/src/offset/local/mod.rs#L63) can't as easily be purely replaced by the internal code, but there is an easy equivalence by using `Local::now().date_naive()`, which does the same thing.

To reduce repetition, I put the error string for `from_ymd` is in a `const`.
For the tests I have been less strict. I added a helper function in the tests for `date.rs` to not make the tests much longer, and in tests for `time.rs` I simply `unwrap()`-ed instead.

Edit: A note is that these aren't deprecated in chrono 0.4.19 as specified in the `Cargo.toml`, but it's nice to not see warnings if a future version is used.

### DIFF
--- a/src/core/date.rs
+++ b/src/core/date.rs
@@ -8,6 +8,8 @@ use chrono::{Datelike, Duration, NaiveDate};
 #[cfg(not(target_arch = "wasm32"))]
 use lazy_static::lazy_static;
 
+const INVALID_DATE: &str = "invalid or out-of-range date";
+
 /// The date value
 #[derive(Clone, Copy, Debug, Default)]
 pub struct Date {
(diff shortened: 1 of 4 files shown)


carl-anders:
Actually nvm, I didn't properly test for chrono 0.4.19. It doesn't have `Local::now().date_naive()`. Closing PR.

PR #86, opened 2023-02-08
Another menu widget

latidoremi:
https://user-images.githubusercontent.com/49757619/217568459-23456f5b-8750-4385-9bb5-6851680dea87.mp4

I did saw the other pr and read their code, didn't quite like it so I wrote one, also I have no idea why the other one seems abandoned.

In the earlier version I tried to put parts of the layout calculation in Widget::on_event(), and output it through the widget state to Overlay::layout(), it makes the most sense to do so bc the layout is dynamic and reacts to events, but iced expects layout before event handling, apart from that, I can't get the viewport size in on_event(), without which there's no way to achieve adaptive open direction

So I went the extra mile to rewrite it so that they are completely separated and the new layout will be calculated after the previous event update. But it's so error prone and causes a ton of weird bugs. I've been trying to clean it up but it's still messy af. Right now it's fairly stable, as least in my tests

I'm quite satisfied with the result tho, it supports any widget, adaptive open direction and path highlight. 

There're still some limitations:
- No adaptive menu width, bc I can't get a minimum required width from any widget, I could have it manually set for each menu, but it's such a hassle I doubt anyone wanna do that.

- No arbitrary item height. It's more of a design choice than limitation, I could support that with flex::resolve(), but when users do want uniform height, which is almost always, they have to manually sync the height of each widget, that's a problem when some widgets don't let you set its height. Maybe it's good to have an option, what do you guys think?

- No scrolling, yet.

Right now the API is a bit verbose, you have to wrap a MenuTree::new() to almost every widget
```
let sub_menu = MenuTree::with_children(
        button("sub menu"),
        vec![
            MenuTree::new(button("item")),
            
        ]
    );
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,6 +100,7 @@ members = [
 #git = "https://github.com/iced-rs/iced.git"
 #rev = "8221794"
 version = "0.7.0"
+features = ["svg"]
 
 [workspace.dependencies.iced_aw]
 path = "./"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,8 @@ wrap = []
 number_input = ["num-traits"]
 selection_list = []
 split = []
+menu = []
+quad = []
 
 default = [
     "badge",
@@ -46,6 +48,8 @@ default = [
     "wrap",
     "selection_list",
     "split",
+    "menu",
+    "quad",
 ]
 
 [dependencies]
@@ -89,6 +93,7 @@ members = [
     "examples/number_input",
     "examples/selection_list",
     "examples/split",
+    "examples/menu"
 ]
 
 [workspace.dependencies.iced]
--- a/examples/menu/Cargo.toml
+++ b/examples/menu/Cargo.toml
@@ -8,5 +8,6 @@ publish = false
 iced_aw = { workspace = true, features = [
     "menu",
     "quad",
+    "icon_text"
 ] }
 iced.workspace = true
--- a/examples/menu/Cargo.toml
+++ b/examples/menu/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "menu"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+iced_aw = { workspace = true, features = [
+    "menu",
+    "quad",
+] }
+iced.workspace = true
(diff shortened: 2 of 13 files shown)


genusistimelord:
It looks good. Do you want it merged now Or would you like me to wait till you add Scroll support and such? Wont merge until you address any review issues I spotted though. Also ignore the Format check it has been erroring for everything. Need to figure out how to get it fixed XD.

latidoremi:
> It looks good. Do you want it merged now Or would you like me to wait till you add Scroll support and such? Wont merge until you address any review issues I spotted though. Also ignore the Format check it has been erroring for everything. Need to figure out how to get it fixed XD.

Actually, I'd like to see more discussions about the things I mentioned, I dont really know what to do with some of them, it's probably not good to make all the decisions myself, and yes if you want it merged, scroll support is going to take a week or more, there wont be breaking api changes

genusistimelord:
menus are a bit hard to deal with and I would not suspect them to work for everything since a menu generally is very limited in what it can support. This is one of the reasons the other was put off for so long. waiting for iced to either add some work around to something that is needed OR to not care and move on with it. For now its a decent looking Menu and just need to get it tested and get scroll support in. I think it will be as good as it will get unless you can ask for or help implement changes to Iced that could solve those issues.

For the height issue You could support both options with a enum that tells the system how to deal with it to statically set it or flex resolve it.  The API itself seems fine. I don't think there is a easier way to really do this unless you plan to make a menu builder macro but even then idk how much easier it would actually make it.  

For the Error proneness just make sure it works as a Whole for however many different widgets you can. We can always keep a list of what it can use correctly or not.

latidoremi:
Just added basic scroll support, might need some more tests, height options on the way. I went back to the original method again lmao, doing layout after event feels way more intuitive

latidoremi:
@genusistimelord I think it's basically done, feature wise. I didn't add flex for heights tho, it seems the resolve function was not meant for usage outside the layout method, there's quite a bit of change needed to implement it in the current system, I'll leave it until it's necessary. Anyway, is there anything else I need to do for it to get merged?

genusistimelord:
just to either tell me why that review doesn't or does need to be changed. Also I did make a few changes in the main repo so you might want to merge those changes in so format will run correctly.

latidoremi:
> just to either tell me why that review doesn't or does need to be changed.

Sorry I don't understand what does 'that review' refer to? I don't see any review conversations

genusistimelord (changes requested):

latidoremi:
Ok fix it in the new commit

genusistimelord:
thank you for your hard work this has been merged.

latidoremi:
my pleasure

PR #92, opened 2023-02-23
Add more anchors for FloatingElement.

IohannRabeson:
Add the anchors North, East, South, West with an example that shows the result.


genusistimelord:
thank you for your hard work. Everything here looks good.

PR #102, opened 2023-03-08
Adaptive root direction

latidoremi:
Make horizontal direction of the root menu adaptive to it's position in the viewport

previous
prev_a

after
after_a

The submenus will follow the root direction so that they're less heaped-up

previous
prev_b

after
after_b


genusistimelord:
thank you.

PR #105, opened 2023-03-13
:sparkles: Add cupertino spinner

bbyler:
**UPDATE**:
I ran `cargo fmt`, as indicated. The PR is larger now. `cargo fmt` reformatted all of the existing examples. There are 10 original files that changed according to this PR. 

Hello! 

Pulling inspiration from Flutter's cross-platform widgets (e.g.  [Flutter Cupertino Widgets](https://docs.flutter.dev/development/ui/widgets/cupertino) ), I'd like to start adding "Cupertino"-style widgets to Iced (iced_aw). 

This PR holds the spinner. I've attached a GIF to show it in action, and the example attached demonstrates how to get started. 

Please let me know if you have any questions/feedback! This has only been tested on my Mac desktop, but it should be fine on any native. This should work for web too.

The spinner is not necessarily "official", but it's quite close, from following the Flutter implementation's constants (which are more official). If you have ideas for how to tweak any pixel/colour to get it just right, feel free to chime in! 

Design
-------

 `CupertinoSpinner`
 
 See
 
 1. [Flutter Activity Indicator](https://github.com/flutter/flutter/blob/0b451b6dfd6de73ff89d89081c33d0f971db1872/packages/flutter/lib/src/cupertino/activity_indicator.dart)
 2. [Flutter Cupertino Widgets](https://docs.flutter.dev/development/ui/widgets/cupertino)
 
 for reference. The Flutter source is used for constants. The implementation for this widget pulls together ideas from:
 
    1. the mainline Clock example
    2. the existing `iced_aw` Spinner
    3. the Flutter Activity Indicator above
    4. the QR Code widget


Cargo.toml
```diff
@@ -55,6 +56,7 @@ default = [
 
 [dependencies]
 num-traits = { version = "0.2.15", optional = true }
+time = { version = "0.3.5", features = ["local-offset"] }
```
bbyler, line 59: I know, another dependency. `time` is used in other examples, and the `local-offset` works nicely for generating "ticks" - the time `millisecond()` method conveniently returns a number between 0 and 1,000. `chrono` appears to be the current time lib, so if there's a similar way to implement the functionality with `chrono`, I can switch.

genusistimelord: I guess take a look at https://docs.rs/chrono/latest/chrono/struct.DateTime.html  and let me know if this allows you to do what you need to do with this addition. Otherwise If it does not Please add time as optional behind a feature.

genusistimelord:
need to fix formatting run cargo fmt on the entire project then save any changes.

PR #106, opened 2023-03-17
:sparkles: Add Cupertino-style switch and buttons

bbyler:
Hello! 

Similar idea to the spinner; below is the widget in action 

Cupertino-style Switch

(The Flutter version: https://api.flutter.dev/flutter/cupertino/CupertinoSwitch-class.html )

Buttons (the image scale is way off but it looks something like): 

Cupertino-style Buttons

(The Flutter version: https://api.flutter.dev/flutter/cupertino/CupertinoButton-class.html )

Also, a "Cupertino-style" alert widget

Cupertino-style Alert


genusistimelord:
@bbyler have you had a chance to look at what i have said?

PR #134, opened 2023-06-08
Format the code

Luni-4:
This PR formats the code fixing the CI

### DIFF
--- a/examples/floating_element/src/main.rs
+++ b/examples/floating_element/src/main.rs
@@ -6,7 +6,7 @@ use iced::{
     Element, Length, Sandbox, Settings, Theme,
 };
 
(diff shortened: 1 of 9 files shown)


genusistimelord:
Thank you, Yeah the issue we currently have is either people never use fmt, or even if they do use fmt its Stable,Beta,Nightly which all format differently and the current checker only looks at standard i believe.

Luni-4:
Can we block a PR up till the code has been formatted? Or simpler, up till the CI passes?

genusistimelord:
personally i need to look at how to fix the CI so it doesn't discredit nightly and beta formatting too. I have had formatting break on me and I use nightly.

PR #135, opened 2023-06-08
Update format.yml and clippy.yml

genusistimelord:
add nightly to the test for fmt.

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -2,10 +2,15 @@ name: Clippy
 on: [push, pull_request]
 jobs:
   all:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        rust: [stable, nightly]
     steps:
       - uses: hecrj/setup-rust-action@v1
         with:
+          rust-version: ${{ matrix.rust }}
           components: clippy
       - uses: actions/checkout@master
       - name: Lint with clippy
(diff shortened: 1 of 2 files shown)


.github/workflows/format.yml
```diff
@@ -2,11 +2,16 @@ name: Format
 on: [push, pull_request]
 jobs:
   all:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
```
Luni-4, line 5: We can use `ubuntu-latest` directly here since it's the only supported os

.github/workflows/format.yml
```diff
@@ -2,11 +2,16 @@ name: Format
...
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
```
Luni-4, line 8: Remove this line if the comment above is going to be accepted

.github/workflows/clippy.yml
```diff
@@ -2,11 +2,16 @@ name: Clippy
 on: [push, pull_request]
 jobs:
   all:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
```
Luni-4, line 5: Same as the first comment

.github/workflows/clippy.yml
```diff
@@ -2,11 +2,16 @@ name: Clippy
...
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
```
Luni-4, line 8: Same as the second comment

Luni-4 (commented):
Just a simple review, I hope it will be appreciated

PR #138, opened 2023-06-11
feat: dark style for tabs

### LINKED ISSUE #125: [Help] Custom style for Tabs
How to implement custom style for `Tabs`?
I'm trying to add a `Dark` style when the current theme is `Dark`
I've tried to do the following:
```rs
#[derive(Debug, Clone, Copy, Eq, PartialEq, Display, Default)]
pub enum MyTheme {
    Light,
    #[default]
    Dark,
}

impl iced_aw::tab_bar::StyleSheet for iced::Theme {
    type Style = MyTheme;

    fn active(&self, style: Self::Style, is_active: bool) -> iced_aw::style::tab_bar::Appearance {
        // ...
    }

    fn hovered(&self, style: Self::Style, is_active: bool) -> iced_aw::style::tab_bar::Appearance {
        // ...
    }
}
```

This does not work due to one of Rust's principles:
__*"Only traits defined in the current crate can be implemented for types defined outside of the crate define and implement a trait or new type instead"*__

If this approach is not feasible, what is the correct one? Thanks 😊🎉

PS: What do you think if we add `Dark` in `TabBarStyles`?
PSPS: I think styling is not strictly related to `Tabs` but also for other widgets too... Add an example on how to override a style? 👀

carlocorradini:
Fix #125 

The colors are exactly the reverse of those in the default (light) style.

### DIFF
--- a/examples/tabs/src/settings.rs
+++ b/examples/tabs/src/settings.rs
@@ -103,7 +103,7 @@ impl Tab for SettingsTab {
                 ))
                 .push(Text::new("TabBar color:").size(20))
                 .push(
-                    (0..5).fold(Column::new().padding(10).spacing(10), |column, id| {
+                    (0..6).fold(Column::new().padding(10).spacing(10), |column, id| {
                         column.push(
                             Radio::new(
                                 predefined_style(id),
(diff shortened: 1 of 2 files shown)


genusistimelord:
thank you.

Issue #133, opened 2023-06-07
Extend modal with right click and display it where is clicked


wiiznokes:
It would be nice to have a widget to do actions with the right click of the mouse.

I think the `menu` widget could serve as a good basis for the implementation, because apart from `MenuBar`, we can use `MenuTree` I think. It would also be necessary to add the detection of the right click.

I will be happy to implement it if I can have some advice :)

genusistimelord:
probably could implement this for modal. the only downside is we also need to keep track of mouse movement and if its within the mouse or not to allow auto closing of it.

wiiznokes:
I'm not sure if modal is suitable.
I meant something like this: 
![Capture d’écran du 2023-06-08 00-49-50](https://github.com/iced-rs/iced_aw/assets/78230769/c393e6e8-9d3a-4e38-add3-4e738e892952)

genusistimelord:
that would be a modal that's activated on right click basically. modal render on top when they are shown. So we can use the modal to make the ability to make right clickable screen actions. we might even be able to render it to the spot we click as well if iced allows this yet.

wiiznokes:
Ok this seems simpler than using `menu`.
> we might even be able to render it to the spot we click as well if iced allows this yet.

the `menu` and `picklist` have this behavior I think, why iced would not allow that ?

```
fn mouse_interaction(
        &self,
        state: &Tree,
        layout: Layout<'_>,
        cursor_position: Point,
        viewport: &Rectangle,
        renderer: &Renderer,
    ) -> mouse::Interaction {
        self.underlay.as_widget().mouse_interaction(
            &state.children[0],
            layout,
            cursor_position,
            viewport,
            renderer,
        )
    }
```
I see there is no filtering in the `mouse_interaction` function of modal. So is the right click disable by default ? because only the left click work for opening it.

genusistimelord:
the current modal uses a underlay to open the modal itself. So we would just need to make a new modal that can interact with right click. But the concepts you need to do this are in modal.

wiiznokes:
Ok I think I get it, I just have to modify `mouse_interaction` to catch a right_click, if not, I call `underlay.mouse_interaction`.
And also find a way to display the overlay in the right place.

I think the new function will look like this: `new(underlay, overlay)`

For the state, I think it must be managed inside the widget, unlike the modal.

I do not understand the usefulness of the struct `State`?

```
#[derive(Debug, Default)]
pub struct State<S> {
[...]

PR #139, opened 2023-06-11
Impl ContextMenu

wiiznokes:
This pull request implement #133.

I think this widget could be improved in the future by adapting menu to be used as overlay (only one tree of course not the entire bar).

The overlay appears at the place where the right click was made but apart from that, the implementation is quit naive and does not take into account the available space. But the menu have already this behaviour.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -328,5 +328,23 @@
             "args": [],
             "cwd": "${workspaceFolder}"
         },
+        {
+            "type": "lldb",
+            "request": "launch",
+            "name": "Debug executable 'context_menu'",
+            "cargo": {
+                "args": [
+                    "build",
+                    "--bin=context_menu",
+                    "--package=context_menu"
+                ],
+                "filter": {
+                    "name": "context_menu",
+                    "kind": "bin"
+                }
+            },
+            "args": [],
+            "cwd": "${workspaceFolder}"
+        },
     ]
 }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ split = []
 menu = []
 quad = []
 spinner = []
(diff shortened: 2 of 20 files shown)


genusistimelord:
Ok so in order for this to work Correctly as a right click anywhere menu you will need to remove the underlay portion from context menu.

wiiznokes:
> Ok so in order for this to work Correctly as a right click anywhere menu

you mean if we combine menu and context menu ?

> you will need to remove the underlay portion from context menu

why ?

> Probably should make the underlay Optional actually as this might be better for those who just want to create a screen right click area

but the event is not propagated through the app. We could make a method `on_right_click` but this will be confusing I think.

And sorry but I don't understand why we need to click over other widgets ?

genusistimelord:
well no I mean atm you are tied to the underlays position for a rigth click window. which is OK if you only want to give options of right click to certain things within a underlay. It would be nice if this could just be added and not count towards the Layout and just take right click events but this would require a special kind of change in iced. but the current method can work. we can merge this once cargo fmt is ran on the source.

wiiznokes:
I'm not sure I have enough knowledge of Iced to understand this approach. Maybe a start of api will help me.
But I think it's already good for a start.

I finished formatting everything.
We agree that there is no auto-fix command for `cargo fmt` right?

genusistimelord:
if you type only "cargo fmt" and nothing else it should format the entire project including anything that was missed.

wiiznokes:
> if you type only "cargo fmt" and nothing else it should format the entire project including anything that was missed.

I guess I just lost 30 minutes :smile:

genusistimelord:
it is a good learning Experience. You could always join the discord and ask for help on these things too.

PR #140, opened 2023-06-22
Remove unnecessary `Clone` requirement on FloatingElement's `Message`

DoomDuck:
Removing this seemingly unnecessary `Clone` attribute allows for `Message`s holding one usage not `Clone` attributes.
Such as a unique handle to a connection for example, or a `channel::Sender`.


genusistimelord (approved):
Looks good thank you for your help.

PR #141, opened 2023-07-07
split: Remove unneeded Message: Clone bound

f4814:
As far as I can see this bound is not needed anywhere and removing it should not break anything


genusistimelord (changes requested):
Please fix the clippy warnings and formatting of  the source.

f4814:
As far as I can see none of the warnings are caused by my commit or am I missing something? AFAICT every warning/error originates from `src/native/cupertino` or `examples/cupertino` which do not use the split.
(Also: Yes I was stupid yesterday and wrote spinner instead of split)

PR #143, opened 2023-07-12
Fix clippy warnings

Luni-4:
This PR fixes clippy warnings for the main library, **not** examples. 

I do not know how to treat the bound repetition warning, so I left that untouched putting an `allow` attribute


Luni-4:
@bbyler 

Do you know how to fix the clippy errors in the `cupertino` examples?

PR #144, opened 2023-07-12
Remove nightly version for clippy and format checks 

Luni-4:
`clippy` and `rustfmt` checks for `nightly` are really unstable, so better to use only the `stable` version


genusistimelord (approved):
Yeah I do agree they have been more broken recently than they have in the past.

Issue #151, opened 2023-07-15
cannot find type `Appearance` in module `tab_bar`


Redhawk18:
```toml
iced_aw = { version = "0.5", default-features = false, features = ["menu", "split", "tab_bar", "tabs"] }
```

Trying to do custom styles I found that i could import `menu` and style it just like native iced widgets, however it seems `tab_bar` doesn't have the appearance trait when import as 
`use iced_aw::{menu, tab_bar};`

this can be fixed by doing `use iced_aw::style::tab_bar;`

However this shouldn't be the case it seems like a oversight to me.

example
```rust
use iced_aw::{menu, tab_bar};

impl tab_bar::StyleSheet for Theme {
    type Style = Theme;

    fn active(&self, style: Self::Style, is_active: bool) -> tab_bar::Appearance {
        Default::default()
    }

    fn hovered(&self, style: Self::Style, is_active: bool) -> tab_bar::Appearance {
        Default::default()
    }
}
```
```
error[E0412]: cannot find type `Appearance` in module `tab_bar`
  --> src/gui/theme.rs:49:71
   |
49 |     fn active(&self, style: Self::Style, is_active: bool) -> tab_bar::Appearance {
   |                                                                       ^^^^^^^^^^ not found in `tab_bar`
   |
help: consider importing one of these items
   |
1  + use crate::gui::button::Appearance;
   |
1  + use crate::gui::text_input::Appearance;
   |
1  + use crate::gui::theme::application::Appearance;
   |
1  + use crate::gui::theme::menu::Appearance;
   |
     and 17 other candidates
help: if you import `Appearance`, refer to it directly
   |
49 -     fn active(&self, style: Self::Style, is_active: bool) -> tab_bar::Appearance {
49 +     fn active(&self, style: Self::Style, is_active: bool) -> Appearance {
   |

error[E0412]: cannot find type `Appearance` in module `tab_bar`
  --> src/gui/theme.rs:53:72
   |
53 |     fn hovered(&self, style: Self::Style, is_active: bool) -> tab_bar::Appearance {
   |                                                                        ^^^^^^^^^^ not found in `tab_bar`
   |
help: consider importing one of these items
   |
1  + use crate::gui::button::Appearance;
   |
1  + use crate::gui::text_input::Appearance;
   |
1  + use crate::gui::theme::application::Appearance;
   |
1  + use crate::gui::theme::menu::Appearance;
   |
     and 17 other candidates
help: if you import `Appearance`, refer to it directly
[...]

PR #152, opened 2023-07-15
fixed `Appearance` not being avaiable through `iced_aw::tab_bar` 

Redhawk18:
#151

Added a import to fix the bug.


genusistimelord (commented):
looks good, Just run "cargo fmt" and then it should be good to merge.

PR #160, opened 2023-07-29
Doc Cleanup

rs017991:
Fixed an outdated Modal example, then noticed that a lot of the other docs were not building correctly on docs.rs.
Ended up doing a full sweep on the docs.


genusistimelord (approved):
looks good thank you for your help.

Issue #158, opened 2023-07-27
Menu Widget ItemHeight/Width Dynamic


genusistimelord:
Would be nice to Add a dynamic ItemHeight and or Width. This would at least allow the menu to adapt to the controls rather than the controls adapting to the menu. easier to setup with less issue of Clipping text etc. @latidoremi

latidoremi:
Dynamic item height is doable, but I'm not so sure with dynamic item width

For dynamic width to work, the minimum width should be available from every item in order to compute a proper width of a menu. `Widget::width` is not useful in this case, if all of the items in a menu have `Length::Fill` widths, we can only fallback to a default width which is not that different than setting a width manually. Even if some of the items have `Length::Fixed` widths, with which we can compute a proper width for them and have other items comply to it, there'll be gaps when their widths are different

<img width="126" alt="gaps" src="https://github.com/iced-rs/iced_aw/assets/49757619/00f74e8d-2e60-42fe-9f79-e7c09f79f1d9">

So unless we have something like `Widget::get_min_size`, I don't think dynamic item width is possible in the current design

Alternatively we can go with Kaiden42's design, where only certain types of item are allowed, practically it's just converting button, checkbox, text, etc. into the menu item version, and we can define a min width for each of them, but that'll be a different menu widget

genusistimelord:
I think with dynamic width the best thing to do is maybe have a maximum size width limit. Also the reason I am suggesting these is due to the nature of how fonts and such work now. Fonts no longer use the Width and height of the Box they are in instead its split up into  Font size and line_height. which messes up the rendering if you set the Height to be statically smaller than the font size *  line height. so it causes clipping issues. I did fix the clipping issues in the example by making the size Bigger, but it would be nice to have a Max and Min limit and then allow Dynamic, static and uniform to be based around these.

latidoremi:
So basically
```
let item_width = match widget_bounds_width{
    Length::Fill => static_width,
    Length::Fixed(x) => x.max(static_width)
};
let item_height = match widget_bounds_height{
    Length::Fill => static_height,
    Length::Fixed(x) => x.max(static_height)
};
```
Is that correct?

But what does max size limit has to do with the clipping issues?

genusistimelord:
that would work but I think the issue with Fill is the Max would need to Clip the Fill so it cant go past a certain Point. Especially for width. Height I'm not to concerned here about but it would be nice to have some overall limitation even if its huge by default.

[...]

PR #163, opened 2023-08-04
Menu dynamic item height

latidoremi:
Add `ItemHeight::Dynamic` #158 

Automatically chooses proper item height in these cases:
- Fixed height
- Shrink height
- Menu tree height

When none of these is the case it will fallback to the default value


genusistimelord (approved):
looks good thank you!

PR #164, opened 2023-08-06
Submenu offset

### LINKED ISSUE #150: Menu items rendering imprecisions
Menu items are rendered imprecisely, indeed submenu entries are not placed exactly under the menu bar, but instead immediately under the texts. I think it would be useful to adopt the same rendering approach used for the `pick_list` widget which shows submenus immediately under the rectangle. The same approach, in my opinion, could be used for the right-arrow submenus.

latidoremi:
Fix #150 by adding `main_offset` and `cross_offset`

`main_offset` moves all the menus in the vertical open direction

`cross_offset` moves each menu in the horizontal open direction

Before
submenu_offset before
<br/>
After (main_offset = 13,  cross_offset = 16)
submenu_offset after

### DIFF
--- a/examples/menu/src/main.rs
+++ b/examples/menu/src/main.rs
@@ -198,6 +198,8 @@ impl Application for App {
         }
         .spacing(4.0)
         .bounds_expand(30)
+        .main_offset(13)
+        .cross_offset(16)
         .path_highlight(Some(PathHighlight::MenuActive))
         .close_condition(CloseCondition {
             leave: true,
--- a/src/native/menu/menu_bar.rs
+++ b/src/native/menu/menu_bar.rs
@@ -65,6 +65,8 @@ where
     spacing: f32,
     padding: Padding,
     bounds_expand: u16,
+    main_offset: i32,
+    cross_offset: i32,
     close_condition: CloseCondition,
     item_width: ItemWidth,
     item_height: ItemHeight,
@@ -89,6 +91,8 @@ where
             spacing: 0.0,
             padding: Padding::ZERO,
             bounds_expand: 15,
+            main_offset: 0,
+            cross_offset: 0,
             close_condition: CloseCondition {
                 leave: true,
                 click_outside: true,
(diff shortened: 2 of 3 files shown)


latidoremi:
Is the naming ok? `main` and `cross` were borrow from flex::Axis, they sound kinda obscure, but I can't think of better ones, `vertical` and `horizontal` is a bit misleading, "moving all" and "moving each (relative to their parents)" is different

latidoremi:
Honestly, I don't even know what "submenu offset" means lol, it's random name came to my mind

genusistimelord:
I think it is fine as long as documentation describes what they are for.

latidoremi:
Can this be merged now? Do you think the docs aren't detailed enough or are there other problems I should look at?

genusistimelord:
> Can this be merged now? Do you think the docs aren't detailed enough or are there other problems I should look at?

I was waiting for you to tell me it was ready. XD

genusistimelord (approved):
everything looks good thank you.

Luni-4:
@latidoremi 

Thanks a lot for your implementation!

Just a question, if I reduce `main_offset` and `cross-offset` to 0, I will get a one-after-another (for vertical menus) and side-by-side (for horizontal menus) behavior, right? Or will a minimal amount of offset always be present?

genusistimelord:
try to see what occurs i guess.

latidoremi:
offset

If you set both to 0, you'll get exactly what it looks like before. The default `background_expand` is [6;4], so you need to set `main_offset`=6, `cross_offset`=12 to get the one-after-another behavior

In a more general case when `background_expand`=[top, right, bottom, left], you need to set `main_offset`=top or bottom, `cross_offset`=2*(right or left), you won't get the one-after-another behavior if either top!=bottom or right!=left

Luni-4:
Thanks a lot @latidoremi for your explanation! 

Can we add this information in a `README.md` into this [directory](https://github.com/iced-rs/iced_aw/tree/main/examples/menu)? In this way, anyone can consult it anytime

latidoremi:
Good idea, I should probably explain all the other parameters as well, I'll do it asap

PR #165, opened 2023-08-09
feature/tab_bar icon glow

Redhawk18:
a glow is added when hovering over the close icon, also fixed a typo.

Peek 2023-08-09 17-55


genusistimelord (approved):
looks good lets see if clippy and format think so too!

PR #167, opened 2023-08-14
refactor improve types and simplified usage 

Redhawk18:
I removed the tuple that the function pointer had, because there was no reason to have it in first place. With that I changed the type of `options` to a slice `&'a [T]` making its so many more types can be passed into to widget at creation. Doing that I had to change the required impl's from `ToString` to `Display`.

Also I'm considering adding support if the enum has a default, having that default be automatically highlighted. lmk if that's a good idea


genusistimelord (approved):
looks good thank you. Originally we had to use Cow for some reason, that I simply can not remember =D or that over time rust itself fixed the issue on the backend.

PR #169, opened 2023-08-14
Extend modal to support multiple overlays

lucatrv:


genusistimelord:
looks good just fix the Clippy issues and we are all set to merge. Was thinking of a Vec instead but this works too.

PR #170, opened 2023-08-16
support normal scrolling instead of inversed scrolling

Redhawk18:


genusistimelord (approved):
looks good, just need to run cargo fmt and should be good to go.

Redhawk18:
for some reason running `cargo fmt` leads to no changes but the workflow is still failing

PR #174, opened 2023-08-31
add multiple modals example

lucatrv:


lucatrv:
@genusistimelord before merging this please see issue #175

genusistimelord:
could you also remove the not needed comment from the documentation.. https://github.com/iced-rs/iced_aw/blob/main/src/native/modal.rs#L68

lucatrv:
@genusistimelord please review the new PR, and let me know if everything is ok or if it should be further improved. Then I will also update the documentation comment, so please do not merge it yet.

genusistimelord:
> @genusistimelord please review the new PR, and let me know if everything is ok or if it should be further improved. Then I will also update the documentation comment, so please do not merge it yet.

looks good.

lucatrv:
@genusistimelord, while updating the documentation comment that you pointed me out I noticed that most of the links of the `iced_aw` documentation have duplicated identical labels. For instance `[Modal](Modal)` which from what I know can be simplified as `[Modal]`. Please let me know if I should:
1. Do not care about that and update only the comment that you pointed me out
2. Update all comments in `modal.rs` removing the unneeded duplicated identical labels
3. Remove the unneeded duplicated identical labels from all `iced_aw` documentation

genusistimelord:
you can fix those too but id prefer it to be in a separate PR if you want to go though and tidy up all the documentation.

lucatrv:
Should I fix that single comment in this PR, and then tidy up all documentation links in another PR?

genusistimelord:
probably if there is that much that needs to be fixed best to do it in the other PR with the rest.

lucatrv:
So you can go ahead merging this, and I will open a separate PR for the documentation.

genusistimelord (approved):
everything looks good thank you.

PR #183, opened 2023-09-11
Card overlays

turai:
Currently cards can't have overlays, which this PR implements, however a few questions remain:
- All the children have their own wrapper layout, I'm sure there's a reason for it. This means the overlay method must be called for the first child within the wrapper (the actual child layout), otherwise only the first one will be rendered (eg. Row [ PickList (this only), PickList, PickList ]). This requires either an unwrap or some kind of logic, which is, even if it's just a line, may not be necessary. Could this be a valid reason to remove the wrappers and use the child layouts direcly?
- is there a point to have overlays for the header or the footer? In my opinion only allowing the body to have overlays could be sensible, however I do have some use cases from the past where I had to use dropdown menus in the footer

### DIFF
--- a/src/native/card.rs
+++ b/src/native/card.rs
@@ -575,8 +575,9 @@ where
                 .zip(&mut tree.children)
                 .zip(layout.children())
                 .filter_map(|((child, state), layout)| {
-                    let first_child = layout.children().next().unwrap();
-                    child.as_widget_mut().overlay(state, first_child, renderer)
+                    layout.children().next().map(|child_layout| {
+                        child.as_widget_mut().overlay(state, child_layout, renderer)
+                    }).unwrap_or(None)
                 })
                 .collect::<Vec<_>>();
 
--- a/src/native/card.rs
+++ b/src/native/card.rs
@@ -561,29 +561,29 @@ where
     }
 
     fn overlay<'b>(
-            &'b mut self,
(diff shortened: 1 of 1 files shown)


src/native/card.rs
```diff
@@ -559,7 +559,30 @@ where
...
+            let mut children = vec![&mut self.head, &mut self.body];
+            if let Some(foot) = &mut self.foot {
+                children.push(foot);
+            }
+            let children = children
+                .into_iter()
+                .zip(&mut tree.children)
+                .zip(layout.children())
+                .filter_map(|((child, state), layout)| {
+                    let first_child = layout.children().next().unwrap();
```
genusistimelord, line 578: I think it should be possible to ignore the panic here by using a If Let Some(first_child) = layout.children().next().
you can make it return a None on an Else and then just filter out all the None's before Collect.

genusistimelord (changes requested):

genusistimelord (dismissed):
looks good now. just need to fix any formatting or clippy issues and I can merge it in.

genusistimelord (approved):
Everything looks good now! Let me know if your done with any changes and I can merge this in.

genusistimelord:
OK that can be in another PR later as its not really super important. this can be the current fix at least.

Issue #178, opened 2023-09-03
Release-event handling in `ContextMenu` 


Remmirad:
When nesting a `Slider` in a `ContextMenu` I get the behavior that when dragging the `Slider` and releasing the mouse button outside the `ContextMenu` and then open it again I am still dragging the `Slider` without clicking it again. 

This is probably caused by `ContextMenuOverlay` only forwarding a `ButtonReleased` when the release happens inside the `ContextMenuOverlay` bounds (see https://github.com/iced-rs/iced_aw/blob/dc18fca47a312f7e7e5857a3ef48b414aaed6199/src/native/overlay/context_menu.rs#L163). 

Maybe this Event (_possibly also include `FingerLifted` here_) could just always be forwarded to the contained element? Then it is up to the contained widgets to decide what they want to do here.

PR #187, opened 2023-09-30
Update event handling for context_menu overlay

Remmirad:
This tries to fix https://github.com/iced-rs/iced_aw/issues/178 .

This changes when events are forwarded to the child of the context menu. 
It allows widgets like sliders to be used in a context menu since it now forwards "Released" events to the child. (_Otherwise the slider would still be grabbed when reopening the menu_)

This also prevents a click event from going _through_ the context menu, when e.g. a "Released" event happens inside the content bounds but the content returns `Ignored`. Since the context_menu takes action here by closing itself it also should catch the event.

### DIFF
--- a/src/native/overlay/context_menu.rs
+++ b/src/native/overlay/context_menu.rs
@@ -144,10 +144,13 @@ where
             .next()
             .expect("Native: Layout should have a content layout.");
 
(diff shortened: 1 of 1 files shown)


genusistimelord (approved):
looks good thank you.

PR #188, opened 2023-10-05
Set tag for MenuBar

dave-wathen:
If MenuBar replaces a different (stateless) widget in the view for an application it causes a downcast error because the state is never initialised correctly: since tag() is not implemented for it, diff() gets called instead of state()/children().


genusistimelord (approved):
looks good thank you.

genusistimelord:
you closed it before i could merge it?

PR #189, opened 2023-10-08
Wrap example is broken

ojji:
The `wrap` example looks like this to me:
broke

Changing the controls column width from the default `iced::Length::Shrink` to either `iced::Length::Fixed(value)` or `iced::Length::FillPortion(value)` fixes the example program.

### DIFF
--- a/examples/wrap/src/main.rs
+++ b/examples/wrap/src/main.rs
@@ -113,7 +113,7 @@ impl Sandbox for RandStrings {
         let RandStrings {
             vbuttons, hbuttons, ..
         } = self;
-        let vertcal = Container::new(
+        let vertical = Container::new(
             vbuttons
                 .iter()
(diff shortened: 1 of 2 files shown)


genusistimelord:
yeah that is weird as iced::Shrink should make it the smallest possible size. So something is definitely not right in column or within the Text portion which is causing the size of the boxes to get Extra big.

ojji:
I agree this is just a band aid. It looks like to me that the `NumberInput` is causing the issue, at least changing the default width in the `NumberInput::new()` function from `Fill` to `Shrink` will stop pushing the controls column all the way to the right.
However i'm not familiar with the library enough to know if that would break other things or not, or is the right solution anyway.

genusistimelord (dismissed):
looks good

genusistimelord (approved):

PR #190, opened 2023-10-11
Implement new grid widget

avsaase:
As discussed on Discord, this PR replaces the grid widget with a new one with more layout options. The updated example shows the layout options and their effect.

https://github.com/iced-rs/iced_aw/assets/880421/9a7e3fe8-32be-45e8-9716-af20953abfc6

I think this is an improvement over the previous implementation but it but the API is geared towards a fixed number of elements per row and thus it does not support a variable number of columns and dynamically wrapping around cells to the next row. 

Some limitations:
- This does not play nice with widgets that try to fill up all available space. For example, when you add a `iced_widget::Row` with `.width(Length::Fill)` to the grid, it will push the grid partially out of the window. I'm not sure how to fix this. I believe the previous version also didn't account for this.
- There's no option to let the grid itself fill the remaining space by widening the columns like in `iced_widget::Row`. I wanted to add this but it requires a new layout algorithm to account for all the edge cases so for now I kept it simple.


genusistimelord (dismissed):
looks good. just need to let clippy and fmt check it and once that passes ill merge it in.

genusistimelord:
for the Width::Fill I think in order for it to work you need to send the Exact size of the box its trying to fill to the Element to get the Elements correct size and layout created.

genusistimelord (dismissed):

avsaase:
> for the Width::Fill I think in order for it to work you need to send the Exact size of the box its trying to fill to the Element to get the Elements correct size and layout created.

But how can I get the size of the box that a child widget wants to fill?

genusistimelord (approved):

avsaase:
Sorry, last commit. I found a fix for the issue where the grid gets pushed around when a child as a `Fill` width.

genusistimelord:
that would probably be a hector question on how is the best way to get the Current boxes size so inside each rows box you can then know what it can be and send that to the element so it cant grow bigger. As even I struggle with the Layout stuff.

genusistimelord (approved):

genusistimelord:
> > for the Width::Fill I think in order for it to work you need to send the Exact size of the box its trying to fill to the Element to get the Elements correct size and layout created.
> 
> But how can I get the size of the box that a child widget wants to fill? I don't fully understand how the layout engine works when widgets try to take more space than the minimum size.

you don't get the inner elements sizes you get the outer elements, So if you know the size of each Column in a row you send that to the element to  tell the element they can not go past this size. generally we should be able to get this size. but that would be a Hector question on discord or his forum board. IF you want me to leave this PR open for a little bit let me know, otherwise I will merge this and whatever new changes can be PR'd later.

avsaase:
I think this can be merged. I need to learn more about the layout algorithm before making more changes.

PR #191, opened 2023-10-20
Implement grid flex layout

avsaase:
This PR implements a new version of the grid widget I added in #190. 

This version has two major improvements over the previous implementation: 

- It support taking up more space than the minimum to fit the contents with `width()` and `height()` methods, similar to the built-in row, column and container widgets.
- Has options to control how column widths and row heights are allocated when the grid is stretched.

The algorithm to allocate space to the rows and columns when the flexible size is enabled is somewhat complex because I wanted the relative columns widths/row heights to be dependent on the relevant fill factors of the length settings. When there's not enough overall space to achieve this ration it's "taken away" first from the columns/rows that have space left given their contents. It's a little hard to describe but the updated example makes it clear.

I added `itertools` as a dependency because it adds some methods on iterators that made implementing much more concise and probably more performant.


genusistimelord (approved):
looks good!

PR #197, opened 2023-11-20
icon fonts

Redhawk18:
* adds the nerd font
* adds a `icon_to_string()` function to make code cleaner
* decouple bootstrap icons from the default


genusistimelord:
not really decoupling it if your still using it internally.  what does the nerd font give btw?

Redhawk18:
> not really decoupling it if your still using it internally. what does the nerd font give btw?

I would say this describes it best
nerd icon

But out of this I'd really like the seti icons (default vscode icons) and the material icons.

I misspoke about decoupling, it is more straight forward what is happening now if you want to use the bootstrap icons vs another icon font.

genusistimelord:
what is the license for it?

Redhawk18:
> what is the license for it?

https://github.com/ryanoasis/nerd-fonts/blob/master/LICENSE

Redhawk18:
since licenses are a issue, I believe using nerd font will be a bad idea. I will add the smaller fonts that make it up.

genusistimelord:
there license is ok.

PR #201, opened 2024-02-02
Update overlay code to compile with iced master

brianch:
Update to make things compile again after [this](https://github.com/iced-rs/iced/pull/2226) iced change.

I'm unfamiliar with how to create widgets in iced, to be honest I changed things following a bit blindly iced's official widgets changes, so please be skeptical of this PR 😅  The widgets appear to be working correctly and the changes are almost all trivial, but I'm still not 100% sure if it doesn't introduce problems in the modal position for example.

### DIFF
--- a/src/native/card.rs
+++ b/src/native/card.rs
@@ -14,7 +14,7 @@ use iced_widget::{
         renderer, touch,
         widget::{Operation, Tree},
         Alignment, Border, Clipboard, Color, Element, Event, Layout, Length, Padding, Pixels,
-        Point, Rectangle, Shadow, Shell, Size, Widget,
+        Point, Rectangle, Shadow, Shell, Size, Vector, Widget
     },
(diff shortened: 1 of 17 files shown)


genusistimelord:
have you run all of the examples to ensure they all run?

also what was the model p[position issues? I already know they are not perfectly centered since the last changes atm just dont know how to go about getting the correct yet.

brianch:
Yes, the examples work (the grid needs an additional change, the checkbox message is now set on the "on_toggle" function, and there's a panic related to glyphon while setting the horizontal alignment to center or right, but seems unrelated).

I mentioned the modal because the position isn't being explicitly set anymore, so I don't know if this will cause issues.

genusistimelord:
could you fix the clippy and formating issue. I can then merge it and take over any needed changes to fix any issues.

PR #209, opened 2024-02-23
Menu rewrite

### LINKED ISSUE #196: Beta Update to latest Iced Master
Currently I have updated everything to the current except for Menu. @latidoremi could you please work on updating this to the latest master branch for me? as it might require some major changes to make it work due to how hector changed Tree.

latidoremi:
Fix #196 

Complete rewrite of the menu widget

`MenuTree` is now divided into `Item` and `Menu`, since widgets have to be wrapped inside menu tree nodes and most items won't have a submenu, wrapping with `Item` only costs each widget one extra pointer

New helper macros: `menu_item`, `menu`, `menu_bar`, details are in the docs

Menu bar background styling

Mouse interaction is now implemented

Rename `PathHighlight` to `DrawPath`, a `FakeHovering` variant is added, details are in the docs

`ItemWidth` and `ItemHeight` are removed, static width and dynamic height should suffice, other options can be easily implemented by the lib users

Todos:
- CloseCondition
- Multioverlay

I'm a bit tired working on this rewrite, I'll come back to the todos later, but I need to take break for now

@genusistimelord can you test it and see if there are bugs I didn't catch, if there are't any you can merge it now

Btw I have to comment out this
```
[workspace]
members = [
    ...
    # "examples/floating_element multioverlay",
    ...
```
for some reason cargo won't compile the menu example if I don't


genusistimelord:
yeah the example Errors because I forgot to upload the folder. I did that just now so it should work now and you need to merge those into this.

genusistimelord:
OK I see a few issues with the Scroller menu.

1: When Selecting the jcsu button the popup menu cuts itself off at the top and only works when it finally detects the jcsu button is paste half way and displays it below.
2: The scrolling of the inner menu causes the button/label to scrunch rather than scroll. 
image
3: the scrollbars are not visible for direct scrolling and it is Very slow to scroll via middle mouse button.


Other than those everything else looks good.

latidoremi:
> OK I see a few issues with the Scroller menu.
> 
> 1: When Selecting the jcsu button the popup menu cuts itself off at the top and only works when it finally detects the jcsu button is paste half way and displays it below. 2: The scrolling of the inner menu causes the button/label to scrunch rather than scroll. image 3: the scrollbars are not visible for direct scrolling and it is Very slow to scroll via middle mouse button.
> 
> You will also need to run cargo fmt and cargo clippy and fix all the issues too.
> 
> Other than those everything else looks good.

1: Fixed

2:This is the expected behavior atm, the algorithm simply clips whatever that is not visible, it does not know where the clipping happens, and it scales the layouts at the two ends to fit the space, one benefit is that it keeps the content in the middle so when a tall widget gets clipped the text stays in the middle. I can fix that but it'll take some time

3:Scroll speed added, scrollbar was not planned, I can do it but again it'll take some time, I have to save it for later

Those fmt and clippy issues come from floating element and tabs, should I fix them?

genusistimelord:
yes please fix the clippy issues and fmt issues.

latidoremi:
Ok on second thought the scaling does look stupid it's now fixed

genusistimelord (approved):

genusistimelord:
thank you for all your hard work =D.  Also thank you for fixing the scaling XD

justDeeevin:
LFG!!

PR #214, opened 2024-02-29
handle unavailable cursor position in TabBar

StillGreen-san:
hover checks in TabBar handle an unavailable cursor position by using a default position.
this default position however is inside the window, resulting in a potential positive hover check if the cursor is outside the window

2024-02-29_18-16-06_NVIDIA_Share 2024-02-29_18-16-14_tabs
_(hovering an active tab will not show a highlight, it is actually 'hovered' in the second screenshot)_

this pull changes the bounds checks in tab_bar to treat an unavailable cursor as false

2024-02-29_18-18-49_NVIDIA_Share

i have to admit that
```rust
if cursor
    .position()
    .map_or(false, |pos| layout.bounds().contains(pos))
```
is a bit harder to read than
```rust
if layout
    .bounds()
    .contains(cursor.position().unwrap_or(Point { x: -1., y: -1. }));
```
_at least for me_

but i choose the former as
- a: im not sure if negative values are actually impossible to be returned by cursor.position()
- b: i think it more accurately represents the intention, of returning when false there is no overlap, and without a cursor position there cant be one. instead of checking an arbitrary invalid point

i also split this into 3 commits _(since i dont know what all of these functions actually do)_
- the first one addresses the "issue" i had
- the third applies the same logic to the remaining checks

at leas on desktop everything seems to still work _(in the TabBar example)_

### DIFF
--- a/src/native/tab_bar.rs
+++ b/src/native/tab_bar.rs
@@ -672,16 +672,15 @@ fn draw_tab<Theme, Renderer>(
         item.expect("Graphics: Layout should have an texts layout for an IconText")
             .bounds()
     }
-    let is_mouse_over = cursor
-        .position()
-        .map_or(false, |pos| layout.bounds().contains(pos));
+
+    let bounds = layout.bounds();
+    let is_mouse_over = cursor.position().map_or(false, |pos| bounds.contains(pos));
     let style = if is_mouse_over {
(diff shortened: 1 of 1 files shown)


genusistimelord (approved):
looks good as well.

PR #215, opened 2024-02-29
change TabBar hovered appearance for default style

StillGreen-san:
im not sure if this was intentional, but hovering over an active tab will not highlight it anymore, which i find a bit odd

2024-02-29_19-15-01_tabs 2024-02-29_18-00-51_tabs

looks like this behavior was introduced in #206

this pull just changes it to always use the strong highlight color, similar to how was before

2024-02-29_18-01-36_NVIDIA_Share


genusistimelord (approved):
looks good thank you

Issue #208, opened 2024-02-23
Date Picker breaks with non-default font or font size


Strosel:
The date picker does not expose any way to change the font, font size or cancel/confirm icons

With a non-default font and font size this is the result
<img width="323" alt="image" src="https://github.com/iced-rs/iced_aw/assets/18191699/f95c8540-d6a2-4311-9483-497f3b820ce0">

It would be beneficial if the date picker widget exposed methods to change the font, font size and maybe even icons used to render the widget. Changing the size of each day in the picker as to not clip the text would also be nice if possible.

Strosel:
Same issue with the close button in`Card`
<img width="215" alt="image" src="https://github.com/iced-rs/iced_aw/assets/18191699/eb1f0967-8ad0-4105-abcf-c5912225995a">

wangeguo:
| Before (Expected)  iced_aw: 0.7.0 | After (Wrong) iced_aw: 0.8.0 |
| ------------- | ------------- |
|![截屏2024-02-26 18 36 56](https://github.com/iced-rs/iced_aw/assets/146697/71eaa0a3-1431-4261-a6aa-5c07083be447)|![截屏2024-02-26 18 36 37](https://github.com/iced-rs/iced_aw/assets/146697/f2fb35ca-af71-45b4-9abc-6e14059b9e1f)|

`BootstrapIcon::ChevronExpand`, `BootstrapIcon::Plus` and `BootstrapIcon::Box`

genusistimelord:
[...]

PR #216, opened 2024-03-03
Add font size to date picker

Strosel:
Adds the ability to set font size on the date picker overlay as discussed in #208.
Also dynamically changes the size of the overlay to best fit the relevant font size.


genusistimelord (approved):
looks good thank you.

PR #219, opened 2024-03-05
reran icon script and removed all garbage unicode values

Redhawk18:
* added `icon_to_text` that will return a iced text widget
* renamed icons to their actual font's names
* removed garbage font names

for who ever has to rerun this script in 2-4 years to update the icons here are the prefix and suffix text files to output the same result.

[prefix.txt](https://github.com/iced-rs/iced_aw/files/14489495/prefix.txt)
[suffix.txt](https://github.com/iced-rs/iced_aw/files/14489493/suffix.txt)


genusistimelord (approved):
looks good.

PR #224, opened 2024-03-17
only activate required iced-features

spamviech:
They are automatically activated for users of this crate.

My specific use-case:
For some reason, starting with iced-0.12 activating the wgpu-feature results in a "unwrap on a None-value" panic when executing on a rasperry pi.
Additionally, I don't want to activate the "tokio"-feature, since I don't use it for anything else (yet), and tokio isn't exactly small.

All examples are still compiling (added missing "icons"-feature where rustc was complaining).
However, I noticed some weirdness (not addressed here, to keep the scope small):
- wgpu requirement in examples:
  - panic on tab_bar (start) and multiple_modals (after button press)
  - sometimes starts with black window
  - color-picker bottom line stays empty
  - menu: when focusing an item, every other item is only a black box
- cuppertino_switch has weird behaviour (e.g. toggling both visually)
- split_scroller
  - panics when used with "wgpu" feature
  - horizontal scrolling is not shown in slidebar at the bottom
- folder for `slidebar` is named `sliderbar` (extra "r" in the middle)
- folder for `widget_id` is named `WidgetIDReturn`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3991,9 +3991,9 @@ checksum = "53be06678ed9e83edb1745eb72efc0bbcd7b5c3c35711a860906aed827a13d61"
 
 [[package]]
 name = "yansi"
-version = "1.0.0-rc.1"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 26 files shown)


genusistimelord (approved):
looks like a reasonable change, thank you

PR #228, opened 2024-04-03
fix: button doesn't highlight when it is hovered

Sherlock-Holo:
this PR can fix the example button doesn't highlight when it is hovered


genusistimelord (approved):
thank you

PR #229, opened 2024-04-06
add missing set_active_tab

spamviech:
Found during an investigation in the same area, the current tab was not highlighted correctly.


genusistimelord (approved):
thank you.

PR #230, opened 2024-04-07
Hide fill_quad behind intersects/>0-checks to avoid triggering debug_assert in the tiny_skia-renderer

spamviech:
This prevents triggering the most common debug_assert-checks in the tiny_skia-renderer (`width/height.is_normal()`).

The intersects-check match the container-widget.
However, since there is no (easy) access to the viewport for overlays use simple >0-checks instead.

Additionally, remove the wgpu-feature from the remaining examples (see https://github.com/iced-rs/iced_aw/pull/224), since they are not crashing anymore.

Notes:
- Sometimes, the window stays black at the start. Usually, resizing the window a tiny amount resolves this (A restart might also help).
- When resizing the window to very small sizes, the renderer might still panic/hit the `debug_assert`. This might be a similar issue in iced itself, since it also happens in e.g. the number_input-example when commenting/removing all (direct) `fill_quad`-calls.

### DIFF
--- a/examples/WidgetIDReturn/Cargo.toml
+++ b/examples/WidgetIDReturn/Cargo.toml
@@ -11,7 +11,5 @@ iced_aw = { workspace = true, features = [
     "number_input",
     "icons",
 ] }
-iced = { workspace = true, features = [
-    "wgpu",
-] }
+iced.workspace=true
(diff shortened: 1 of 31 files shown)


genusistimelord (approved):
looks good thank you.

PR #233, opened 2024-04-19
Number_input fixes

### LINKED ISSUE #232: Numberinput requires mouse hover for keyboard input
I got a [bug report](https://github.com/the-marenga/sf-scrapbook-helper/issues/67), that the number inputs in my application do not work.
The issue seems to be, that numberinput requires the mouse to be above the widget, even if the number inside is currently selected (blinking cursor) in order to accept keyboard inputs. This is not intuitive and not how the regular text_input behaves, which leads to users thinking they only work via the arrows. 

To make sure this is not a bug in my code, I ran the number_input example in this repo from the main branch and the issue is there as well.

My questions now are:
Is this a bug, or the desired behaviour? 
Did I miss any option to toggle this behaviour?
Am I doing something wrong?

the-marenga:
Sorry in advance for turning a one line fix into this, but I just kept finding unrelated bugs, while I was checking if the initial fix had any bugs and just kept fixing them.

These fixes include:
- The initial fix for having the text input only trigger on mouse hover
- Each number input would trigger two messages to be emitted, instead of just one. The first was explicitly triggered by `shell.publish` and the second one was from the text input parsing the same event one line later via `self.content.on_event`, which triggered it's on_input function. 
- The selection replace/delete logic had an off by one error, which just resulted in the replacement being skipped, if the selection contained the last character. The seconds parsing of the event by the textinput would then however parse the selection “correctly” and send a different message with a different value, than the first. As that value would be emitted later, it would overwrite the first value
- This leads to any selection replacement of the last character to be accepted. Meaning, that values > max, or < min could be set, as every max would effectively have a 9 and a min of 0 as the last digit no matter what the actual max value is
- Similarly negative max bounds, could have their value set to zero, by selecting and deleting the current value
- The delete key could be used to set any value, by either deleting the ‘-’ char to turn negative numbers positive, or deleting the ‘.’ to turn fractions into large numbers, as these inputs were not checked at all

I was losing track of where which event part was handled, so I repackaged all of it into mainly one match and only pass keyboard events to the textinput, that are correctly checked

To make the testing for myself easier, I also added the ability to use ctra+a,v,c & accept '-' as a key input

There is still some minor jank around negative numbers and bounds that do not contain 0, but I am pretty sure these things could only be solved by not passing the event to the child/text for inserts/deletes, which would require reimplementing a bunch of selection & cursor movement logic

Closes #232

### DIFF
--- a/src/widgets/number_input.rs
+++ b/src/widgets/number_input.rs
@@ -397,200 +397,164 @@ where
             .next()
             .expect("fail to get decreate mod layout")
             .bounds();
-        let mouse_over_inc = inc_bounds.contains(cursor.position().unwrap_or_default());
-        let mouse_over_dec = dec_bounds.contains(cursor.position().unwrap_or_default());
-        let modifiers = state.state.downcast_mut::<ModifierState>();
-        let child = &mut state.children[0];
 
         if self.bounds.0 == self.bounds.1 {
             return event::Status::Ignored;
         }
 
-        if layout
-            .bounds()
-            .contains(cursor.position().unwrap_or_default())
-        {
-            if mouse_over_inc || mouse_over_dec {
-                let mut event_status = event::Status::Captured;
-                match event {
-                    Event::Mouse(mouse::Event::ButtonPressed(mouse::Button::Left)) => {
-                        if mouse_over_dec {
-                            modifiers.decrease_pressed = true;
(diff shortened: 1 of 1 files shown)


genusistimelord (approved):
looks good thank you.

PR #236, opened 2024-04-27
Fix number input over/underflow

### LINKED ISSUE #235: Number input doesn't prevent overflow and underflow of the contained value
If a number input contains a value such that

```rust
self.value - self.step
```

underflows, then the value underflows (in release mode) or the program panics (in debug mode.) This behavior can easily be replicated in main by changing the "number_input" example to take u32, putting 1 as a default value, 0 as a minimum, and 2 as the step.

The same thing happens with overflows.

I made a fix locally and I am going to do a PR to propose it upstream.

TitouanReal:
Fixes #235.

This code change makes sure to not do any operations that might overflow or underflow.

The code makes the assumption that it does not receive the widget in a broken state, ie the invariant

```rust
self.bounds.0 <= self.value <= self.bounds.1
```

must be true for the code to be correct. If it is, then the result after the operation will uphold the same invariant.

Such a topic is very error prone, so please feel free to try and find problems in the code I wrote.

### DIFF
--- a/src/widgets/number_input.rs
+++ b/src/widgets/number_input.rs
@@ -227,7 +227,7 @@ where
         if self.bounds.1 - self.bounds.0 < self.step || self.value < self.bounds.0 + self.step {
             self.value = self.bounds.0;
         } else {
-            self.value = self.value - self.step;
+            self.value -= self.step;
         }
         shell.publish((self.on_change)(self.value));
     }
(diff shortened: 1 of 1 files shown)


genusistimelord (dismissed):
looks good thank you.

genusistimelord:
fix the clippy issue and you should be good.

TitouanReal:
It should be fixed now.

genusistimelord (approved):

PR #237, opened 2024-04-30
Replace Length with impl Into<Length> in width/height APIs

TitouanReal:
Widgets in iced allow to change their height and their width with an API like this :

```rust
pub fn width(mut self, width: impl Into<Length>) -> Self
```

This is especially useful when setting a fixed width/height:

```rust
my_widget.width(30);
```

instead of 

```rust
my_widget.width(Length::Fixed(30.));
```

This PR proposes to change the API to make it more in line with the one in iced.

### DIFF
--- a/src/widgets/badge.rs
+++ b/src/widgets/badge.rs
@@ -94,8 +94,8 @@ where
 
     /// Sets the height of the [`Badge`].
     #[must_use]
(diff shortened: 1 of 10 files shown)


genusistimelord (approved):
thank you, this was still set to the old way of doing things.

Issue #223, opened 2024-03-10
A way to disable on scroll behavior in number_input


Exidex:
If you have a number input in scrollable() area and you scroll when your mouse just happened to be on top of the input you will accidentally change the value of input which is not desirable. So I ask for a boolean property on NumberInput to disable on scroll function. 

I also would argue that no on scroll behavior should be the default.

Redhawk18:
Sounds like a easy first pr

Exidex:
Resolved by https://github.com/iced-rs/iced_aw/pull/239

PR #239, opened 2024-05-07
Allow disabling scroll events for NumberInput.

airblast-dev:
https://github.com/iced-rs/iced_aw/issues/223

Simply allows disabling mouse scroll events for the `NumberInput`. Not sure how needed this is, but I can see it being annoying where the outer scrollable is small, as the user could accidentally modify the value.

The issue also mentions the idea of the event being opt-in, instead of opt-out which can be changed in `NumberInput::new` if preferred. Though personally I don't think its worth changing default behavior.


genusistimelord (approved):
looks good

PR #242, opened 2024-05-14
changed number input's bounds to an inclusive range

Redhawk18:


Redhawk18:
I was thinking about it and theres a lot of small changes I want aswell

genusistimelord (commented):
I would like you to change this to use RangeBounds So it can use any of the Range types like
..
0..
0..100
..100
0..=100

I would also like you to add an additional trait requirement. https://docs.rs/num-traits/latest/num_traits/bounds/trait.Bounded.html

This will allow use to get a min and a max for all the Generics. Which for the .. style bounds we can then set the minimal and maximum numbers to the Unbounded.

genusistimelord (dismissed):
This looks much better. Thank you.

Redhawk18:
@genusistimelord

genusistimelord (approved):
approved

PR #254, opened 2024-07-03
Use IntoIterator instead of Vec for tabs

barskern:
I have an app where I create the list of tabs at runtime through the use of an iterator. The allocation of a Vec which is then immediately iterated upon seemed a bit wasteful. If there was a reason for the particular API, please do tell. If not, here is a simple change which changes the function to accept any IntoIterator, which would then avoid this allocation in my case. Further, the existing interface should be preserved, as Vec implements IntoIterator, in addition to that the size_hit will return the length of the Vec.

Thank you for a great couple of widgets in this library!

### DIFF
--- a/src/widgets/tabs.rs
+++ b/src/widgets/tabs.rs
@@ -115,15 +115,18 @@ where
     ///     * the function that will be called if a tab is selected by the user.
     ///         It takes the index of the selected tab.
     pub fn new_with_tabs<F>(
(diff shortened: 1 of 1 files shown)


genusistimelord (approved):
Looks good, it should be this. I have not really focused on updating iced_aw much, since most of the widgets are made by other people. I just try to keep it updated to iced itself.

PR #257, opened 2024-07-13
Toggle Menu Bar on Click

sundaram123krishnan:
## Requirement

In our [pigg](https://github.com/andrewdavidmackenzie/pigg) app, we encountered a scenario where it was necessary to close the menu bar when the user clicks on the menu bar itself.

## Implementation
To address this need, I implemented the feature to close the menu bar when clicked. When the menu bar is clicked, it toggles its open/close state.

### Example:

Any feedback is appreciated!

### DIFF
--- a/src/widgets/menu/menu_bar.rs
+++ b/src/widgets/menu/menu_bar.rs
@@ -223,7 +223,7 @@ where
             }
             Event::Mouse(mouse::Event::ButtonReleased(mouse::Button::Left)) => {
                 if cursor.is_over(bar_bounds) && bar.is_pressed {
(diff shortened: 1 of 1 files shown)


genusistimelord (commented):
I think this is Fine as long as it does not interrupt drag and drop and so forth of inner widgets. Could you please verify the change works correctly with the drag color bars in the example?

sundaram123krishnan:
I just checked now, it definitely works fine

genusistimelord (approved):
looks good then thank you.

sundaram123krishnan:
Thx

PR #258, opened 2024-07-13
use new alignment methods from iced

Redhawk18:
I had a few compile errors, I assume because I pulled a newer version of Iced. So here's a commit that fixes those build errors.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "ab_glyph"
-version = "0.2.26"
+version = "0.2.28"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


genusistimelord:
yeah looks like you have a bunch of errors you need to fix as well.

genusistimelord:
hmm just noticed this was for the Beta Branch. The Beta branch was already merged into the maain branch a while ago. So if you could sync this over that way it would be much nicer.

Redhawk18:
> hmm just noticed this was for the Beta Branch. The Beta branch was already merged into the maain branch a while ago. So if you could sync this over that way it would be much nicer.

There was a new breaking update from iced, Thats why the lock file got updated.

PR #263, opened 2024-07-18
New feature: Sidebar

rizzen-yazston:
Similar to the `tabs` and `tab_bar` features, though the new feature `sidebar` implements the tabs in a vertical column on the side of the tab content, instead of horizontal row above or below as done for `tabs` and `tab_bar`.

Also contains two additional widgets `FlushColumn` and `FlushRow`.

There is a single line correct in `tab_bar` feature.


genusistimelord (approved):
looks good thank you.

PR #264, opened 2024-07-18
Remove errors occuring when sidebar feature isn't enabled

43-21:
The types Sidebar and SidebarWithContent didn't have cfg attributes.


rizzen-yazston:
thanks for catching that one. miss adding the required cfg attributes

genusistimelord (approved):
thank you for catching this!

PR #265, opened 2024-07-19
Corrected issue in 'layout()' of 'FlushColumn' and 'FlushRow'.

rizzen-yazston:
the copy and paste gremlin. fixed it. now `Sidebar` works as expected.


genusistimelord (commented):
thank you for your Fix but you need to fix the clippy warnings before I will merge.

genusistimelord (approved):
your changes meet requirements despite the cargo update that broke selection list.

PR #267, opened 2024-07-19
Adding support for dots in number_input

### LINKED ISSUE #234: number_input, unable to type non-integer numbers.
Good afternoon.
With the number_input, I am unable to type non-integer numbers. On the example provided for use of the widget, I can obtain `10.5` for example if I press the arrow up or scroll.
2024-04-26-155931_grim

However, if I try to type it, I am unable to.
2024-04-26-155925_grim

I think the cause is that the dot is parsed and removed before I can type what goes after it. Furthermore, I can also input a non-integer number if I type it elsewhere and copy and paste it into the application.

I don't know wether I have to enable a setting I am unaware of or if this is unimplemented functionality.

Thank you for everything.
Kind regards.

Ultraxime:
Should fix #234 

To fix it, I created a new widget, the TypedInput, it allows the user to write anything in a TextInput, but only send messages when it is a correct value, and then the message contain this value (not the string).

I then modify NumberInput to use this widget instead of a standard TextInput.

My code may not be the cleanest or respect all the guideline of the repo, so I'll be happy to edit it according to your liking.

### DIFF
--- a/src/widgets/number_input.rs
+++ b/src/widgets/number_input.rs
@@ -503,15 +503,15 @@ where
                             if new_val.is_empty() {
                                 new_val = T::zero().to_string();
                             }
-
+                            
                             match T::from_str(&new_val) {
                                 Ok(val)
-                                    if (self.min..self.max).contains(&val) && val != self.value =>
+                                    if  val >= self.min && val <= self.max && val != self.value =>
                                 {
                                     self.value = val;
                                     forward_to_text(event, shell, child, clipboard)
                                 }
-                                Ok(val) if (self.min..self.max).contains(&val) => {
+                                Ok(val) if val >= self.min && val <= self.max => {
                                     forward_to_text(event, shell, child, clipboard)
                                 }
                                 Ok(_) => event::Status::Captured,
@@ -547,12 +547,12 @@ where
 
                             match T::from_str(&new_val) {
                                 Ok(val)
-                                    if (self.min..self.max).contains(&val) && val != self.value =>
+                                    if val >= self.min && val <= self.max && val != self.value =>
                                 {
                                     self.value = val;
                                     forward_to_text(event, shell, child, clipboard)
                                 }
-                                Ok(val) if (self.min..self.max).contains(&val) => {
+                                Ok(val) if val >= self.min && val <= self.max => {
                                     forward_to_text(event, shell, child, clipboard)
                                 }
                                 Ok(_) => event::Status::Captured,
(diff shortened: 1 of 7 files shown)


Ultraxime:
I'll fix the pipeline once I have feed back (to prevent doing such a small commit)

genusistimelord (commented):
It looks good, but I'd need to test it first before pulling it in.

Ultraxime:
It could be interesting to make TypedInput able to have a different style when the entry is invalid, but I don't understand well enough this part of iced to do it

genusistimelord:
Ok this is a decent update that does fix the issue. However, there was one issue I noticed that needs to be fixed before I can merge this.  When using an u32 and setting the range from 0..250 I should be able to type 249, but I can not and can only type up to 248. Please fix this issue then I can merge it.

Ultraxime:
I think I know where it comes from (using `self.min..self.max` instead `self.min..=self.max`)
But I'm not sure if there is a real purpose to using two fields to store the min and the max, instead of using a field to store something implementing RangeBounds ?

genusistimelord:
if Max is ..250 then it should go to 249 if max is ..=250 then it should go to 250. ATM I set it as ..250 so it should allow typing in 249 but I am only getting 248 as a max to type in.

Ultraxime:
I agree with what you are saying

My question was, wouldn't be better to store the bound using a type implementing RangeBounds.
Or is it a reason to using two fields ?

genusistimelord:
the reason is to allow all Range's to work with this so the end user can dictate how they want their min and max's to be like.

Ultraxime:
Isn't what RangeBounds supposed to work ?

src/widgets/number_input.rs
```diff
@@ -503,15 +503,15 @@ where
                             if new_val.is_empty() {
                                 new_val = T::zero().to_string();
                             }
-
+                            
                             match T::from_str(&new_val) {
                                 Ok(val)
-                                    if (self.min..self.max).contains(&val) && val != self.value =>
```
genusistimelord: Oh i see what you mean now, I thought you meant change how number input takes in the range. Sorry. 
Yeah the ..= would have worked fine here as a range bound lookup. However I think the if val >= self.min && val <= self.max => { is probably a faster method of dealing with it.

genusistimelord (approved):

genusistimelord:
you just need to resolve the format issue by running cargo fmt

PR #272, opened 2024-07-22
add lucide font

Redhawk18:


Redhawk18:
If you could rerun the ci

genusistimelord:
I am not really sure if we should allow adding more fonts onto iced_aw unless they have a specific use. For example, bootstrap is actually used in the widgets. As adding these will just keep bulking down the project.

Redhawk18:
> I am not really sure if we should allow adding more fonts onto iced_aw unless they have a specific use. For example, bootstrap is actually used in the widgets. As adding these will just keep bulking down the project.

That's true, I was requested on discord to add these. But it might not be an amazing idea.

genusistimelord:
Would be better to make a new Font crate for Iced. One that just contains all the different fonts.

Redhawk18:
> Would be better to make a new Font crate for Iced. One that just contains all the different fonts.

Thats prob a better idea.

genusistimelord:
someone needs to make a iced_fonts and place any needed extra font there.

PR #277, opened 2024-08-13
Changing the TypedInput APi to a one similar to the TextInput's

Ultraxime:
I modified the method on the TypedInput to match the ones on TextInput
And change the on_submit message to send a result, either Ok(value) or an error with the current text when it's not valid


Ultraxime:
Normally it's finished 
But I'd be happy to modify things if I missed some 

And I don't know if you think it would be useful to have the same overload of the draw and layout method that Textinput has

genusistimelord:
What you did is good, I was just making sure you were done. I will merge this.

PR #281, opened 2024-08-26
Replacing f32 by iced::Padding and iced::Pixels when possible

Ultraxime:
I replaced the `f32 by `impl Into<Padding>` or `impl Into<Pixels>` in function where it made sense, to mimic the behavior of iced.

And at the same time, replaced the `f32` by `Padding` and `Pixels` in structures to allow more flexibility.

Normally, I tried every example to check that it was correct. (by the way, wouldn't be cleaner to use the example feature of cargo rather than workspace members ?)


genusistimelord (approved):
Looks like a job well done. Thank you.

PR #282, opened 2024-08-27
Reorganising 

Ultraxime:
I moved the examples from being members of the workspace to being examples (so to use it we have to run `cargo run --example <example>`) which seems more cargo-esque.
I updated `.vscode/launch.json` to match these modifications and previous ones that were not considered.

I also renamed the module `widgets` to `widget` to match the organization of iced.

I corrected some errors in the documentation.

I added a pre-commit file, to ease the checking before pushing modification.

If you have any ideas for other things that could be done, I'll be happy to do them.
I plan on making another pull request later to improve the documentation and tests.


genusistimelord (approved):
This looks good, thank you.

If you wanted to go through all the issues and see which ones you might be able to solve, I'd be much appreciative of it.

Issue #283, opened 2024-09-02
Card uses Limits for both header and Content


lanastara:
Hi,

when I try to put a card into a container that restricts the cards height then the card extends outside the parent container by the height of the header.

After some debugging I found the following lines I think might be the issue:
https://github.com/iced-rs/iced_aw/blob/0f616fe3a74b1587b94e9fb49a29c52bb93ef01c/src/widget/card.rs#L260-L278

im my example [here](https://github.com/lanastara/card_error) I set the outer containers height to 100.0 
after layout I get nodes that are
head: 51px
body: 100px

leading to the card using 151px and overlapping with the next card

PR #284, opened 2024-09-02
Shring card body limits by head and foot size

lanastara:
I used a similar logic to iced_core::layout::next_to_each_other just for height instead of width.

I had to move the footer layouting above the body layouting to ensure the card border gets rendered correctly.

fixes: #283


genusistimelord (approved):
thank you.

PR #285, opened 2024-09-03
fix max width being reduced by padding in number_input

StillGreen-san:
this prevents the max width of the number_input widget from being reduced by the padding

in this example we have a row with 3 collumns filled with buttons _(set to fill)_ and the number_input _(with 22 padding)_ in the middle.
left is before, right with this change
2024-09-03-18-18_number_input 2024-09-03-18-18_number_input (2)

_the call to shrink may have been used here to apply padding to the content, but the padding is already set on the content itself so this is not needed_


genusistimelord (approved):
thank you

PR #290, opened 2024-09-18
bump iced to v0.13.0

edwloef:
Resolves #289. Maybe creating a new release on crates.io would be worth considering?

PR #291, opened 2024-09-18
Versioning table update in README.md

Delphinidae84:
Updated the version table according to https://github.com/iced-rs/iced_aw/pull/290. Will chnage the `iced_aw` version once the crate is out there.


genusistimelord:
yeah, ill be doing this simple task myself when I make the version changes. Just bear with me for a bit as I'm doing final testing.
Thank you though.

PR #294, opened 2024-10-14
Make `selection_list` follow iced theme

TheSpartaPT:
Previously `selection_list` had hardcoded colors which were more geared towards light themes and didn't adjust them according to the theme you set for iced, this PR makes it so `selection_list` uses palette of the set iced theme.


genusistimelord (approved):
thank you

PR #297, opened 2024-10-24
Fix compilation errors with 2 examples that use icons.

hydra:
Compilation looks like it was broken by 7d14e42d117a76a2e20bd88bf29117bd4745d2a3

Icons were moved to `iced_fonts` according to the changelog, but it looks like these two examples were missed.

Note: It appears that neither `sidebar` or `tab_bar` examples are compiled when running `cargo build --examples`.

I also note there's no `Build: Passing/Failed` badge for this repo or automated github actions which would likely have highlighted this compilation error earlier.

### DIFF
--- a/examples/side_bar/main.rs
+++ b/examples/side_bar/main.rs
@@ -45,9 +45,9 @@ fn main() -> iced::Result {
         TabBarExample::update,
         TabBarExample::view,
     )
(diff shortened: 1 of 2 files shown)


genusistimelord (approved):
thank you! good catch.

PR #302, opened 2024-11-15
Fixing the compilation error caused by std Instant on WASM targets

whizsid:
I did the changes to only `spinner.rs`. But `cargo fmt` changed the other files.


genusistimelord (approved):
thank you.

PR #303, opened 2024-11-17
Fixing the backspace issue in number inputs on Mac

whizsid:
Mac keyboards are using the ‘delete’ button as the backspace for text inputs. So the unicode character is `\u{7f}`. But in other platforms, it is `\u{8}`. `iced_core::keyboard` module is already handling it. So we can re-use that functionality for backspace.


genusistimelord (approved):
thank you.

Issue #301, opened 2024-11-14
Unwrap in menu overlay causes app to crash


andrewdavidmackenzie:
In my app, some hardware events cause additions to / removals from menus asynchronously.

It seems that when that happens and the menu is open, a crash happens.

It seems to be this Option unwrap:

https://github.com/iced-rs/iced_aw/blob/main/src/widget/menu/menu_bar_overlay.rs#L90

genusistimelord:
We may need to remove the unwrap and replace it with a full if let statement for menu. that should fix the error.

andrewdavidmackenzie:
"Thar she blows...." https://github.com/iced-rs/iced_aw/pull/307

I fixed the 2 or 3 that were causing my particular crash, plus others I could see in the file.

To fix completely might require some restructuring of those methods.

Idea goal would be to remove the allow of unwraps at the top of the file.

If you review the PR with comments, I am willing to address those comments and get the PR into a state for merging.

PR #307, opened 2024-11-27
Avoid a number of crashes due to menu changes while menu is open.

andrewdavidmackenzie:
Fixes https://github.com/iced-rs/iced_aw/issues/301

### DIFF
--- a/src/widget/menu/menu_bar_overlay.rs
+++ b/src/widget/menu/menu_bar_overlay.rs
@@ -87,47 +87,48 @@ where
             parent_direction: (Direction, Direction),
             viewport: &Rectangle,
         ) {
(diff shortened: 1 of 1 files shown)


genusistimelord (approved):
Looks like a decent way to fix these issues.
Thank you and yes I always accept PR's that are properly handled.

andrewdavidmackenzie:
Ok, great, you're welcome.

I'll watch out for a release, so I can stop using my fork in my app.

genusistimelord:
you can use the main git now since its merged. But if you see or Feel something needs to be fixed and have a potential solution to it always feel free to submit a PR for it. I do go through and review all the changes before i accept any of them.

Issue #286, opened 2024-09-04
NumberInput surprising/wrong behaviour with width & content_width


StillGreen-san:
the `number_input` widget has two functions for setting the width (`width` & `content_width`). and as far as i call tell its also the only widget that has `content_width`

setting none or both _(of `width` & `content_width`)_ to `Fill` works as expected, but using only one of them can result in _(imo)_ surprising/wrong behaviour

as an example having two elements in a `row`, a `number_input` and a `button`(that is set to `Fill`):

```rust
let txt_minute = number_input(self.value, 0..999999, Message::NumInpChanged)
    .style(number_input::number_input::primary)
    // .width(Length::Fill)
    // .content_width(Length::Fill)
    .padding(0);
let btn_btn = Button::new(Text::new("Button"))
    .width(Length::Fill)
    .height(Length::Fill);
```

```rust
let content = Row::new()
    .push(txt_minute)
    .push(btn_btn)
    .align_y(Vertical::Center);
```

left: default width; right: setting only `width` to `Fill`; they work as exepcted

![2024-09-03-20-08_number_input_row_none](https://github.com/user-attachments/assets/f022a53f-183c-4857-aa73-14b245e58998) ![2024-09-03-20-09_number_input_row_width](https://github.com/user-attachments/assets/1e6a8f54-cd7c-44e6-a056-5f414659386e)

but if they each are in a `column` (together in a `row`)

```rust
let content = Row::new()
    .push(Column::new().push(txt_minute))
    .push(Column::new().push(btn_btn))
    .align_y(Vertical::Center);
```

then the `number_input` will only render as if not set to `Fill`, but with the column as `Fill`

![2024-09-03-20-11_number_input_rowcol_width](https://github.com/user-attachments/assets/429082b6-d73c-406a-aa1b-ffe296c87825)

and setting only `content_width` to `Fill` will take up all space _(in both row and rowcol layouts)_

![2024-09-03-20-09_number_input_row_content](https://github.com/user-attachments/assets/2c7ca0ee-88f3-4cdb-b803-01f5fbd6b3e2)

___

im also wondering why there are both `width` and `content_width` functions, other settings like `padding` only have one and set the property on both the `number_input` and the `text_input` content.

genusistimelord:
It's really hard to determine the correct path here, especially since iced keeps changing how Fill and Shrink work. I believe he is finally leaving those alone, but before his last change this use to work fine.

StillGreen-san:
[...]

Issue #308, opened 2024-11-28
Error in typed_input


Ultraxime:
Hi,

@genusistimelord, you pushed some new modification to `typed_input` with commits 55d23d88 and 1bed9d4.

With commit 55d23d88, you seemed to have followed some queries in issues like #306 that I explained why they were irrelevant to add, like having the `on_change` returning a result.
For example, now the `number_input` does not work.

With the commit 1bed9d4, you seemed to fix issue #299. In this commit you change the behavior of the `typed_input` when the text finishes with a `.` which changes the generality of the widget. This modification should have been done in `number_input`.

Is there a reason behind this modification?

genusistimelord:
Well we do want typed input to output errors for the change as this is relevant to how it should work since we are dealing with Text to Values and since we support both integers and floats this is indeed needed. Also the . issues could not be fully resolved in number input itself so i had to do it in typed input. I did run the number input and it was working from the  Example just fine? so i don't know what you mean by no longer working?

Also number input should know when Typed input doesn't process something so it can do its own work.  These are just steps to making it more Relevant to taking any input and allowing it to return if its needed or not to the end user. However I do seem to have a issue where the formatting of Strings removes the . due to anything .0 being reduced down to the main number via formatting rounding.  Was thinking the only way to fix that would be to  handle each number to character manually rather than rely on formatting.  Which would be better as we could better control the input too.

genusistimelord:
Ok so after your fixes the only 2 issues I noticed were

The layout messing up and not shrinking the typed input to be before the Arrow buttons. 
![image](https://github.com/user-attachments/assets/7ed4dc6c-d30a-4a54-a93a-1accb594cd36)

and when typing in all 99999 after the input one turned into an 8
![image](https://github.com/user-attachments/assets/4881387a-2aa1-4751-b1ed-bfa49d93dc3c)

though this could be a format issue of some kind maybe. though it always seems to be the 5th number after the decimal that does it. regardless of what the main number is.



The one issue I did not check is if they tried to fit the number input into a Lower height Zone which should Move the buttons down to a single line and set the box to a minimal height for Text size and buttons.

Ultraxime:
Are the arrow buttons supposed to be after the typed input? I thought it was meant to be above the typed input 

For the eight, the issue comes from the fact that float 32 goes from 249.9999847412109375 to 250 and that the formatting rounds it to 249.99998, and after that anything you add can't make it closer to any other number

[...]

Issue #304, opened 2024-11-24
numeric input cannot enter negative numbers


Adam-Ladd:
Numeric input supports negative number types (e.g. i128) however, as it is not possible to enter a minus sign in the numeric input it is functionally impossible to use them. Please add support for entering a minus sign for negative numbers.

genusistimelord:
It can add a negative sign to numbers, But you must first have a number in there before you can click and add a negative sign ATM.  Ill look into what we can do to allow typing it wholly.

genusistimelord:
this should now work fully with the latest main. Please give it a test,

Issue #299, opened 2024-10-26
Can't edit NumberInput using keyboard


ales-tsurko:
I can select the text, can move the cursor, and can add the numbers after the coma, but can't remove or edit the value further.

On the screenshot, for example, I entered a float number, by adding the dot and number after it. But then I can't remove it using the delete/backspace button.

<img width="165" alt="Screenshot 2024-10-26 at 23 50 12" src="https://github.com/user-attachments/assets/e752ade2-1ac6-4d07-87aa-e2154d80dd07">

genusistimelord:
where you doing this from a Mac?

ales-tsurko:
Why did you close it?

ales-tsurko:
> you never replied back. Also I did make some changes in Main to allow entering the Decimal and - without the need for Text. I can reopen it if it is still an issue within the main branch.

Just tested it and now I can edit the number as described. But now it’s sometimes impossible to enter dot/comma. For example, if you run the example for the number input, try to enter three digit and then type dot/comma. Or sometimes it happens with a single digit. In overall, you can now edit a float after it entered, but it has issues entering it.

For your question about where I’m running it: it’s the latest macOS and the number input example.

genusistimelord:
yeah I am trying to solve this issue right now with the enter issue. It has to do with Text Versus the Value not wanting to accept each other and keep the .

genusistimelord:
that seems to be the one thing I can not figure out the best way to Fix ATM. the reason seems to be when the Value is converted to a String if its a 0.0 or a 1.0 it somehow becomes a 1 or 0 as it gets rid of the .0. Not sure how to Deal with this atm.

PR #309, opened 2024-12-03
Correcting several issues with `NumberInput`

Ultraxime:
I rewrote the logic of `NumberInput` to solve issues #308, #304, #299 and the problem with included and excluded bounds.

The `NumberInput` unlike the `TypedInput` never allows invalid entries.
People who try to implement a floating-point number should follow the way the `f32` is done; it can end with a `.`

I also added the support for any characters in the number, it just has to always be a valid number, regarding the `T::from_str` function.

I merge `content_width` and `width`. It seems to fix issues #286 and #278

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.6.0
+    rev: v5.0.0
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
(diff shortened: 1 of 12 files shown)


genusistimelord (approved):
thank you very much! I looked over it and the changes look good.

Ultraxime:
Normally I checked with the examples 
And with one of my projects so it should be working 
But I'm still waiting for feedback

genusistimelord:
I will fully test them tomorrow to see if I can find anymore issues and let you know. Other than that the best thing I can also do is ask those who had issues to try the master branch.

PR #310, opened 2024-12-03
Updating the methods of NumberInput and TypedInput

Ultraxime:
I add all the relevant methods of TextInput in NumberInput and TypedInput

I did not add the secure method for NumberInput as I thought it was not relevant


genusistimelord (approved):
Thank you.

The Secure method makes no sense here as its a number or Type input and not a password input.

PR #313, opened 2024-12-31
Avoid index out of range panic

andrewdavidmackenzie:
When menu/sub-menu changes due to clicking on a menu item, sometimes an index out of range error occurs at line 108, and this avoids it at least.


genusistimelord (approved):
looks good thank you.

PR #317, opened 2025-01-24
Add map support to GridRow

mytdragon:
I was in a situation I was returning a `GridRow` with a different `Message` type so I needed to map them to the type I needed. Since `GridRow` is basically a `Vec<Element>`, I just iterate the elements and call `Element::map` function.


genusistimelord (approved):
ill accept this change. looks good.

genusistimelord (approved):
thank you for formatting it.

PR #325, opened 2025-03-02
iced 0.14.0-dev tab bar

Redhawk18:
tab bar + a search and replace to help with the breaking of `Widget.on_event` to `Widget.update`


Redhawk18:
it looks like cargo fmt also reformatted the entire library

.gitignore
```diff
@@ -3,3 +3,5 @@ pkg/
 .idea
 .DS_Store
 .vscode/settings.json
+.devenv
+.direnv
```
Zarthus, line 7: You should probably commit these in your `~/.config/git/ignore` or `~/.gitignore` (alongside with .envrc), your global ignore file, rather than a project-level ignore file.

Cargo.toml
```diff
@@ -38,57 +38,58 @@ drop_down = []
...
 num-traits = { version = "0.2.19", optional = true }
 web-time = "1.1.0"
 
 [dependencies.iced]
-#git = "https://github.com/iced-rs/iced.git"
-#rev = "b474a2b7a763dcde6a377cb409001a7b5285ee8d"
-version = "0.13.1"
+version = "0.14.0-dev"
+git = "https://github.com/iced-rs/iced.git"
+# rev = "beddf49c"
```
Zarthus, line 73: you could try branch = main too

Cargo.toml
```diff
@@ -38,57 +38,58 @@ drop_down = []
...
 default-features = false
-features = ["advanced"]
+features = ["advanced", "wgpu"]
```
Zarthus, line 75: why does a lib depend on wgpu?

Zarthus (changes requested):
there's too many changes in this  MR tbh, you should probably mark it as draft and undo some of it.

Cargo.toml
```diff
@@ -5,15 +5,15 @@ authors = [
     "Kaiden42 <gitlab@tinysn.com>",
     "Andrew Wheeler <genusistimelord@gmail.com>",
 ]
-edition = "2021"
+edition = "2024"
```
Zarthus, line 8: the MSRV and edition shift probably belong in separate MRs or at least separate commits

flake.lock
genusistimelord, line 1: this should be git ignored

flake.nix
genusistimelord, line 1: this should be git ignored

src/widget/tab_bar.rs
```diff
@@ -569,12 +568,12 @@ where
                                     |on_close| (on_close)(self.tab_indices[new_selected].clone()),
                                 ),
                         );
-                        return event::Status::Captured;
+                        return ();
```
genusistimelord, line 571: so hector removed the status check return.....

Cargo.toml
genusistimelord: I prefer you to comment out version and rev and add branch = "main"

Cargo.toml
genusistimelord: please move this back to dev dependencies it is not needed in the lib itself.

genusistimelord (changes requested):
I request you make some of the changes I stated first. Also I want you to run Each of the Examples to ensure they all work correctly with how they use to function.

Redhawk18:
@Zarthus How about you make those changes yourself, this isn't a MPL2 licensed project I am not obligated to commit improvements

Issue #300, opened 2024-11-10
DropDown out of view if item is in scrollable


JL710:
I displayed some stuff in a scrollable. I added a dropdown for each item in the list that adds the option to do edits. It opens a dropdown left next to the item.

It works fine if the item is always visible. But if it is a item that I first need to scroll to, the popup is at first visible and for each item it gets further down until it is invisible.

I do not have a minimal code example at the moment.

genusistimelord:
ahh yeah, probably need to figure out how to add in the offsets of scrolling. IF you want to try and Tackle the issue We accept PR's

spantaleev:
I'm also experiencing this issue.

------------

I've managed to make things a little better (but not quite perfect), by manually adjusting the `DropDown`'s offset to match the scrollable's position.

Here's some sample pseudo-like code (not quite accurate, since the dropdown lives in widgets that are *inside* the `scrollable` and the `scroll_offset_y` state value needs to be passed down the widget hierarchy), but you hopefully get the idea:

```rust
# .....

let scrollable = scrollable(widget_list);
    .on_scroll(|viewport| {
        Message::Scroll(viewport)
    });

# .....

match message {
    Message::Scroll(viewport) => {
        self.state.scroll_offset_y = viewport.absolute_offset().y;
    }
}

let mut dropdown = build_dropdown();
dropdown = dropdown.offset(iced_aw::drop_down::Offset::from(-scroll_offset_y + 5.0));
```
-----------

With this workaround:

- ✅ The dropdown is nicely positioned 5px below the `underlay` button that triggers the dropdown
- ❌ For dropdown-trigger buttons which require scrolling to be visible, the button that triggers the dropdown cannot be clicked to close the dropdown anymore. Clicking anywhere else makes the dropdown close, but the button itself not working is problematic

I have not figured out how to solve the latter. It's quite a **bad usability issue**, so **despite half the trouble being worked-around here, dropdowns are still not usable in a scrollable**.

JL710:
Interestingly that does not happen with the default tooltip widget.

JL710:
I seems that it is this line (https://github.com/iced-rs/iced/blob/master/widget/src/tooltip.rs#L291) that is different in the tooltip that causes the issue in the drop down (https://github.com/iced-rs/iced_aw/blob/main/src/widget/drop_down.rs#L217).

PR #326, opened 2025-03-02
drop_down overlay translation bugfix for #300

JL710:
This PR fixes https://github.com/iced-rs/iced_aw/issues/300.

The overlay was not correctly positioned. I took a look at the code and found the solution in the tooltip window as described here https://github.com/iced-rs/iced_aw/issues/300#issuecomment-2692750200.

Since this is a bugfix and not a new feature I think it would be a good idea to get the fix in a patch crates.io release soon.

### DIFF
--- a/src/widget/drop_down.rs
+++ b/src/widget/drop_down.rs
@@ -214,7 +214,7 @@ where
             &self.alignment,
             &self.offset,
             layout.bounds(),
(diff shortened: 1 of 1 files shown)


genusistimelord (approved):
looks good thank you for taking the time to look into it.

PR #329, opened 2025-03-11
Fix `Grid` columns inheriting columns' widths to their left

SFBdragon:
This PR changed the `Grid` element's layout behavior.
- Previously, the `Grid` would impose the width requirement of each column to the left onto the next columns.
- This PR allows columns to have a width less than a column to their left.

I believe this was not intentional and is a bug.


genusistimelord (approved):
looks good thank you.

PR #336, opened 2025-06-07
Preserve ColorPicker state until it is reopened or externally modified

### LINKED ISSUE #330: Color Picker is resetting to initial color after short time
Hi, 

I am working on an application, which uses the DatePicker, TimePicker and the ColorPicker. I noticed, that when I click the color button and open the ColorPicker, I can change the colors, but after a short time, the displayed color resets to the initial color provided in the ColorPicker. When I change any widget in the ColorPicker dialog with the mouse and hold the mouse button, the value stays. When I release the mouse button after some short time the color is reset to the initial color. If I use the cross in the square to select a color or change the hue below the square, the indicators (the cross in the square or the black line in the hue) stay on the location of the selected position, but the color bars on the right side and the displayed hex value are reset. I hope it can be seen in the attached picture, the cross has selected a dark color, nearly black, but the right side is displaying full red.

This is a bit weird, since the DatePicker and TimePicker do not behave that way. 

The application does run a subscription, which updates data every second in the background, but that does not determine the color to be used in the ColorPicker (imagine that the application is plotting realtime charts with data coming in every second, then with a button a event dialog is added where a new event can be added which has a timestamp and  a color and a few other properties, so the event color is completely independent on the running subscription). Still I noticed, that the ColorPicker seems to reset the color when the subscription gets new data (so at the 1 second intervals). 

Any idea what could be the cause of it? It is really strange, that the DatePicker and TimePicker don't do this. Also, the example in the examples directory works perfectly fine. This is why I suspect the subscription to have some kind of side effect, but I am not sure.

I basically do the same in the code as for the TimePicker and DatePicker, the button to open the ColorPicker (which is the underlay) are put into a GridRow and then into a Grid.

```
        let color_button =
            Button::new(fa_icon_solid("palette")).on_press(Message::NewEventColorButton);

        let colorpicker = color_picker(
            self.show_color_picker,
            rgba_to_color(&event.color()),
            color_button,
            Message::NewEventColorCancel,
            Message::NewEventColorSubmit,
        );

        let row3 = GridRow::new()
            .push(text("Color:"))
            .push(text(rgb_to_hex(event.color())).width(Length::Fill))
            .push(colorpicker);
```

So any idea what I am doing wrong? Or is this a bug?

Image
Image

sterliakov:
Fixes #330. The `color` value maintained by the widget overlay will be retained until one of the following happens:

* Picker is reopened
* Picker is rendered with initial `color` value different from previous render

This means that any background subscriptions are safe, but explicitly changing the color while the overlay is open still works as intended and changes the color.

Screencast in progress, will post in comments - IDK how to add videos from GH CLI

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,7 +103,7 @@ required-features = ["card"]
 
 [[example]]
 name = "color_picker"
-required-features = ["color_picker"]
+required-features = ["color_picker", "iced/async-std"]
 
 [[example]]
 name = "date_picker"
--- a/examples/color_picker.rs
+++ b/examples/color_picker.rs
@@ -16,6 +16,7 @@ fn main() -> iced::Result {
         ColorPickerExample::view,
     )
     .font(iced_fonts::REQUIRED_FONT_BYTES)
+    .subscription(ColorPickerExample::subscription)
     .run()
 }
 
@@ -25,6 +26,7 @@ enum Message {
     ChooseColor,
     SubmitColor(Color),
     CancelColor,
+    Tick,
 }
 
 #[derive(Debug)]
@@ -54,6 +56,7 @@ impl ColorPickerExample {
             Message::CancelColor => {
                 self.show_picker = false;
             }
+            Message::Tick => {}
         }
     }
 
@@ -80,5 +83,13 @@ impl ColorPickerExample {
             .center_x(Length::Fill)
             .center_y(Length::Fill)
             .into()
+    }
+
+    fn subscription(&self) -> iced::Subscription<Message> {
+        // We're running this subscription in background to see if any unrelated
+        // rerenders can affect the implementation negatively.
+        // You do not need a background subscription to use color picker, this
+        // is simply test code.
+        iced::time::every(std::time::Duration::from_secs(1)).map(|_| Message::Tick)
     }
 }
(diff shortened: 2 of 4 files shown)


sterliakov:
Example with subscription (from this PR, unmodified, there's a subscription rerendering every second)

https://github.com/user-attachments/assets/7d5024ff-b4fd-4ed7-9e8f-1e6cf2a1846b

Example with scheduled external update (diff below)

https://github.com/user-attachments/assets/fb7b4bb4-73e7-4549-91c4-af5e6a9ca948

```diff
diff --git a/examples/color_picker.rs b/examples/color_picker.rs
index 13b20da..634fa79 100644
--- a/examples/color_picker.rs
+++ b/examples/color_picker.rs
@@ -56,7 +56,11 @@ impl ColorPickerExample {
             Message::CancelColor => {
                 self.show_picker = false;
             }
-            Message::Tick => {}
+            Message::Tick => {
+                let now = std::time::SystemTime::now();
+                let now = now.duration_since(std::time::UNIX_EPOCH).unwrap().as_secs();
+                self.color.r = ((now * 16) % 255) as f32 / 255.0;
+            }
         }
     }
 
@@ -90,6 +94,6 @@ impl ColorPickerExample {
         // rerenders can affect the implementation negatively.
         // You do not need a background subscription to use color picker, this
         // is simply test code.
-        iced::time::every(std::time::Duration::from_secs(1)).map(|_| Message::Tick)
+        iced::time::every(std::time::Duration::from_secs(2)).map(|_| Message::Tick)
     }
 }
```

PR #337, opened 2025-06-07
Fix negative inputs and numpad handling by NumberInput

### LINKED ISSUE #261: Keyboard not working with NumberInput
Hello,

I'm working with iced_aw 0.9.3 and iced 0.12.1 and experiencing an issue with the NumberInput.
I can't write inside using the numpad or upper line of my keyboard; whether it's focused or if the mouse is over it.

I even tried to fork the repository and run the example and, I came to the same issue.

So before trying to fix it myself, I was wondering if it's my fault or not.
I'm running it on Archlinux with Rustc 1.79.0.

The issue also occurred on a computer running Windows. 

Thanks in advance

### LINKED ISSUE #333: Issues with Negative Number Handling in number_input
**Description:**  
I'm encountering a couple of issues when trying to allow negative numbers in the `number_input` component.

**Issues:**

1. **Cannot Enter "-" as the First Character:**  
   When trying to type a negative number, the input field does not allow the "-" character as the first character entered.

2. **Erasing Negative Numbers Incorrectly:**  
   Numbers -0 through -9 cannot be erased from the back at all.

**Expected Behavior:**  
- The "-" character should be allowed as the first typed character in the field.
- It should be possible to erase numbers -0 through -9 using the backspace key (or equivalent) without additional steps.

**Additional Context:**  
Please let me know if you need further information or additional reproduction steps. I appreciate your help in resolving this issue.

**Code:**
```rust
use iced::{
    widget::{Container, Row, Text},
    Alignment, Element, Length,
};
use iced_aw::number_input;

#[derive(Default, Debug)]
pub struct NumberInputDemo {
    value: f32,
}

#[derive(Debug, Clone)]
pub enum Message {
    NumInpChanged(f32),
    NumInpSubmitted,
}

fn main() -> iced::Result {
    iced::application(
        "Number Input example",
        NumberInputDemo::update,
        NumberInputDemo::view,
    )
    .window_size(iced::Size {
        width: 250.0,
        height: 200.0,
    })
    .run()
}

impl NumberInputDemo {
    fn update(&mut self, message: self::Message) {
        match message {
            Message::NumInpChanged(val) => {
                println!("Value changed to {:?}", val);
                self.value = val;
            }
            Message::NumInpSubmitted => {
                println!("Value submitted");
            }
        }
    }

    fn view(&self) -> Element<Message> {
        let lb_minute = Text::new("Number Input:");
        let txt_minute = number_input(&self.value, f32::MIN..=f32::MAX, Message::NumInpChanged)
            .on_submit(Message::NumInpSubmitted)
            .step(0.1);

        Container::new(
            Row::new()
                .spacing(10)
                .align_y(Alignment::Center)
                .push(lb_minute)
                .push(txt_minute),
        )
        .width(Length::Fill)
        .height(Length::Fill)
        .center_x(Length::Fill)
        .center_y(Length::Fill)
        .into()
    }
}
```

sterliakov:
Fixes #261. Fixes #333. 

* Allow lone minus sign in input, mapping it to 0 - the same way as for empty input.
* Fix too eager handling of numpad keys - numpad keys are misdetected by core framework (see https://github.com/iced-rs/iced/pull/2278), this PR aligns their behaviour with `TextInput`.

### DIFF
--- a/examples/number_input.rs
+++ b/examples/number_input.rs
@@ -48,7 +48,7 @@ impl NumberInputDemo {
 
     fn view(&self) -> Element<Message> {
         let lb_minute = Text::new("Number Input:");
-        let txt_minute = number_input(&self.value, -10..=10, Message::NumInpChanged)
+        let txt_minute = number_input(&self.value, -100..=100, Message::NumInpChanged)
             .style(number_input::number_input::primary)
             .on_submit(Message::NumInpSubmitted)
             .step(1);
--- a/examples/number_input.rs
+++ b/examples/number_input.rs
@@ -10,12 +10,12 @@ use iced_aw::number_input;
 
 #[derive(Default, Debug)]
 pub struct NumberInputDemo {
-    value: u8,
+    value: i8,
 }
 
 #[derive(Debug, Clone)]
 pub enum Message {
-    NumInpChanged(u8),
+    NumInpChanged(i8),
     NumInpSubmitted,
 }
 
@@ -48,7 +48,7 @@ impl NumberInputDemo {
 
     fn view(&self) -> Element<Message> {
         let lb_minute = Text::new("Number Input:");
-        let txt_minute = number_input(&self.value, 0..=10, Message::NumInpChanged)
+        let txt_minute = number_input(&self.value, -10..=10, Message::NumInpChanged)
             .style(number_input::number_input::primary)
             .on_submit(Message::NumInpSubmitted)
             .step(1);
--- a/examples/number_input.rs
+++ b/examples/number_input.rs
@@ -37,7 +37,7 @@ impl NumberInputDemo {
     fn update(&mut self, message: self::Message) {
         match message {
             Message::NumInpChanged(val) => {
-                println!("Value changed to {:?}", val);
+                println!("Value changed to {val:?}");
                 self.value = val;
             }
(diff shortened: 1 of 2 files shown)


sterliakov:
https://github.com/user-attachments/assets/478f494f-db33-424c-b9da-e79194a2c422

genusistimelord (commented):
these changes do look like they would work. Does this also work for floating point numbers? Also the Video says its corrupted not sure if thats a github issue or not XD. Anyways I am request ultra to take a look at this.

sterliakov:
Hm, the video renders fine on my end now. Perhaps a github glitch?

This works for FP numbers no worse than before. Naive implementation (just slapping an `f64` as an underlying type) won't work anyway, so it should use some numeric wrapper that can be losslessly converted to/from decimal representation. But FP is difficult anyway - remember there are +0 and -0, NaNs, infinities...

sterliakov:
Given that `"5."` should parse as an `f64`, you'll likely be able to type floating numbers just fine. If you have a reference float wrapper code, we can check that too

Ultraxime:
Looks good, I still need to check it.

It's not exactly what I was planning but should be enough while waiting to do something better

genusistimelord:
ok once you approve it I can merge it.

PR #343, opened 2025-07-17
add CustomLayout widget

JL710:
This PR adds a widget that acts as a container but allows the user to define the layouting of its children. Basically a shorthand for a container widget with a replaceable layout function


genusistimelord (approved):
looks good thank you

PR #344, opened 2025-07-20
Fix menu_bar! macro from needing the user of the macro to import menu::Item

fieryhenry:
This is a simple fix to the `menu_bar!` macro, the full path of the `Item` struct was not specified in the macro definition.

This leads to a `use of undeclared type Item` compile error when using the macro and not manually importing `menu::Item` like so: `use iced_aw::menu::Item;`

This just adds the prefix `$crate::menu::` before `Item` in the macro like what is done for the usage of the `menu::MenuBar` struct.


genusistimelord (approved):
thank you for the fix!

PR #347, opened 2025-07-26
chore: build with iced fork and fix for wasm32 builds

NickelWenzel:
* Adapting to the newer iced version introduced bugs which are not fixed here yet
* Number input seems to work
* Menu is buggy
* Everything else hasn't been tested yet

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "ab_glyph"
-version = "0.2.29"
+version = "0.2.31"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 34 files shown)


genusistimelord:
closed it because it is not finished?

NickelWenzel:
I was a little inattentive, this PR was meant to go to my personal fork of `iced_aw` which is in turn is based on my personal fork of `iced`.

It seems the API will change quite a bit with v0.14 and I especially didn't manage to update the `menu_bar` to work correctly (as a little heads-up in case you don't this is not already on your radar), but for my purposes it works "good enough" but nothing to go onto the official repo here.

genusistimelord:
ok thank you for letting me know.

PR #348, opened 2025-07-27
0.14 init with wip Number input

Redhawk18:
Basic requirements for starting the process of updating the crate.

* My font crate got rewrote and I included the new way to use it.
* Deprecated the grid module since Iced has one now
* No widgets are included by default now. I know we've talked about this before and had the default enable all widgets. However I've seen at least 5 different people in the discord complain that they can't use master, because they don't know that they can just disable the non-updated widgets they don't use. This way **hopefully** we get more contributors upgrading only the widget or two they need lowering our amount of work.
* I tried to get number input working, I could power my way all the way through but I don't really understand the changes asked by the compiler so I didn't want to create bugs for someone else to fix. I'd rather just let someone else finish it correctly.


Redhawk18:
well I fucked up git, fuck git

genusistimelord (approved):

PR #349, opened 2025-08-07
Fix badge status

mytdragon:
- the current update didn't have the hover status working, so I added it like built-in iced's widgets
- the snapping was set arbritarily to false instead of using default like built-in widgets
- there were some debug comments that probably should have been removed?


genusistimelord (approved):
looks good thank you

PR #350, opened 2025-08-07
Update tabs/tab_bar to iced 0.14

mytdragon:


mytdragon:
@Redhawk18 on my fork I didn't have the update with iced_fonts. On the tab_bar widgets, I'm not so sure with the 

```rs
let (_content, font, _shaping) = cancel();
// ...
(self.font.unwrap_or(font), self.icon_size)
```
Around `src/widget/tab_bar.rs:693`. It works but might be incorrect so feel free to edit to the correct way if it's wrong :)

genusistimelord (approved):
thank you very much for your help.

genusistimelord (approved):

Redhawk18:
Why are you not using content or shaping?

genusistimelord:
not sure he can since it can use different fonts and icons. so it might have to stay advanced or be expanded to Require they use your font macros to generate the needed stuff.

mytdragon:
The original widget needed a fallback font for the icon, before it was using REQUIRED_FONT but since it's removed, I needed to use something. From what I see, icon content is set at Tab level when the icon font is at TabBar level. Maybe it should just fallback to default font

Redhawk18:
> The original widget needed a fallback font for the icon, before it was using REQUIRED_FONT but since it's removed, I needed to use something. From what I see, icon content is set at Tab level when the icon font is at TabBar level. Maybe it should just fallback to default font

Oh you don't need that then, iced aw just owns that font now. You'll find it at the crates root now in lib.rs

PR #353, opened 2025-08-15
perf: remove unnecessary redraws

faervan:
@genusistimelord your #64e4bb9 added some unnecessary redraws that cause an insane amount of CPU usage even when idle hovering over the menu
Thus I rebased your changes onto mine
also I removed the development notes that were still left when you merged my previous PR


genusistimelord (approved):
looks good thank you.

PR #354, opened 2025-08-17
Various fixes on drop_down's overlay position

mytdragon:
- Add example for testing position of the overlay.
- Now the overlay will be positioned correctly when the dropdown is inside a horizontal scrollable.
- Previously, on top/right of window the overlay would be half cut by the window, doesn't happen anymore.
- Previously, it would crop the overlay to be at the position around the button, and sometimes it wouldn't crop but hover the button. Now it will always try to keep minimum size of the overlay even if it is hovering the button.

An improvement would be to add a PositionStrategy that would:
- Absolute: what is on this PR, aligning where user want but trying to avoid cropping and allow hovering button.
- Strict: would be what sometimes the previous fix was doing, cropping the overlay to keep it around the button at the alignment wanted.
- Smart: would prioritise the Alignment but if it would crop/hover button, try to do in priority: 1. find alignment where there is room,  2. keep alignment but hover button/crop

The bug mentioned it doesn't fix is that when inside a scrollable that scrolled (reproducible on example if you scroll vertically), after opening the dropdown, clicking on the bottom will close + re-open. Only happens inside scrollable.


genusistimelord (approved):
thank you very much for your hard work!

PR #360, opened 2025-09-13
Menu fixes and updates

latidoremi:
### This PR includes:
- A proper implementation for close-on-click
- A proper implementation for nested overlay
- A fix for button appearance changing incorrectly upon clicking
- Scrolling support for `MenuBar`
- Improved macros that support expressions
- New documentation for `MenuBar`
- Debug logs
- Other fixes and refactors

### Issues introduced by several previous PRs are now fixed
- #257: Implements close-on-click the wrong way, it closes the menu bar without cleaning up all the states.

- #271 and #279: Effectively duct tape solutions for the issues caused by #257.

- #353: The changes doesn't make sense. 
It added a `MenuState.active_submenu` field alongside the existing `MenuState.active` field, then replaced all `MenuState.active` usage with `MenuState.active_submenu`, which did not address the high CPU usage issue.

### Debug logs

I have added debug logs using the `log` crate. It's frustrating having to reinsert all the debug logs whenever a new issue comes. These logs have zero performance cost when the `debug_log` feature is off. Can remove if not desired.

### API changes

- `check_bounds_width` is now `safe_bounds_margin`.

- The new macro syntax now requires a comma after each item.
    ```rust
    menu_items!(
        (widget),       // creates an Item::new(widget)
        (widget, menu), // creates an Item::with_menu(widget, menu)
        expression,     // any expression that returns an Item
        // ...
    )
    ```
- `bar_background_expand` and `menu_background_expand` are removed from `Style`,
`menu_background_expand` is now `Menu.padding`

### DIFF
--- a/examples/menu.rs
+++ b/examples/menu.rs
@@ -644,17 +644,17 @@ impl App {
         ];
 
         let sc = scrollable(c)
-        // .direction(scrollable::Direction::Vertical(
-        //     scrollable::Scrollbar::default()
-        // ));
-        .direction(scrollable::Direction::Both {
-            vertical: scrollable::Scrollbar::default(),
-            horizontal: scrollable::Scrollbar::default(),
-        }); 
+            // .direction(scrollable::Direction::Vertical(
+            //     scrollable::Scrollbar::default()
+            // ));
+            .direction(scrollable::Direction::Both {
+                vertical: scrollable::Scrollbar::default(),
+                horizontal: scrollable::Scrollbar::default(),
+            });
         /*.direction(scrollable::Direction::Both {
-                vertical: scrollable::Properties::new().alignment(scrollable::Alignment::End),
-                horizontal: scrollable::Properties::new(),
-            });*/
+            vertical: scrollable::Properties::new().alignment(scrollable::Alignment::End),
+            horizontal: scrollable::Properties::new(),
+        });*/
 
         fn back_style(theme: &iced::Theme) -> container::Style {
             container::Style {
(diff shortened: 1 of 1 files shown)


Ultraxime:
It looks like a good work, I haven't checked all yet.

I agree on your remarks on logging, adding log is a good idea and leaving them too (depending on the amount and pertinence) 
However, I'm not sure that conditional compilation is necessary, isn't the log crate design to not cost a thing when the log is not meant to be displayed.
Also, wouldn't it be better to use tracing than log as this is meant to be an asynchronous crate?

latidoremi:
From the documentation of the `log` crate:
> If no logging implementation is selected, the facade falls back to a “noop” implementation that ignores all log messages. The overhead in this case is very small - just an integer load, comparison and jump.

If I don't gate them behind a feature flag, and if the users don't set a max level flag themselves, there would be a tiny but non zero perf cost from the log macros. In practice they are very negligible, but I'm not sure if that's acceptable for the repo, personally I'd prefer zero perf cost from logging out of the box

I find `log` good enough for my use case, but if the repo prefers `tracing` I can switch over no prob

genusistimelord:
> From the documentation of the `log` crate:
> 
> > If no logging implementation is selected, the facade falls back to a “noop” implementation that ignores all log messages. The overhead in this case is very small - just an integer load, comparison and jump.
> 
> If I don't gate them behind a feature flag, and if the users don't set a max level flag themselves, there would be a tiny but non zero perf cost from the log macros. In practice they are very negligible, but I'm not sure if that's acceptable for the repo, personally I'd prefer zero perf cost from logging out of the box
> 
> I find `log` good enough for my use case, but if the repo prefers `tracing` I can switch over no prob

I think Log is a fine addition as long as its locked behind a feature. As even the NOOP can get really Wild if the functions are called a lot so to avoid this locking them behind a feature makes more sense to 100% turn it off when not needed.

genusistimelord (approved):
looks good for me. Lots of changes took a while to look through them all XD.

PR #379, opened 2025-12-27
Expose Card close button through operate() method

tonybierman:
## Problem

The Card widget's close button was rendered manually and not exposed as a widget in the operate() method. This prevented iced_test from finding and interacting with the close button, making it impossible to write proper integration tests for close button functionality.

## Solution

Converted the close button from manual rendering to an actual Button widget:

### Widget Structure Changes:
- Added close_button: Option<Element> field to Card struct
- Created create_close_button() method to build a styled Button widget
- Updated on_close() to create and store the button widget

### Widget Tree Integration:
- Updated children() to include close button in widget tree
- Updated diff() to handle all combinations of foot/close_button
- Updated layout() to properly position the close button widget
- Updated update() to delegate button events to the Button widget
- Updated operate() to expose close button (enables iced_test interaction)
- Updated draw() to render the button widget instead of manual icon rendering

### Cleanup:
  - Removed is_mouse_over_close field (Button now handles hover state)
  - Removed unused imports

##  Testing

```bash
# Unit tests
cargo test --lib widget::card

# Integration tests
cargo test --test card_integration_tests
```

## Impact Assessment:

For typical usage with standard iced types (iced_widget::Theme and iced_widget::Renderer), this change should be a non-breaking change.  However, for users with custom themes or renderers, this could cause compilation failures.

## References

[Iced RFC 7 -Widget Operations](https://github.com/iced-rs/rfcs/blob/master/text/0007-widget-operations.md)


genusistimelord (approved):
thank you, yeah the plan was to go through and change as much as possible to be Actual Widgets rather than manually rendered which would simplify the widgets internals. Just at the time Hector did not think it would work fully yet. though his recent changes probably fully allow it to work.

Issue #377, opened 2025-12-27
Color Picker, improve direction


Ultraxime:
Hi,

Using `color_picker` in one of my projects, some users gave me feedback.
They would find it useful if:
 - We could input the hex value of the color directly 
 - We would use a `NumberInput` instead of `Text` for the value of each rgba

Is it a good idea to implement this, and is it even possible?
I'm still not at ease with the way it is implemented.

genusistimelord:
these should be possible and are actually one of the few changes I wanted to make to many of the existing widgets. To switch manual out for widgets, where possible.

Ultraxime:
Ok, 
So if I have the motivation, I could totally rewrite the colour picker widget for it to use widgets internally and not some strange behaviour?

ghenry:
> Hi,
> 
> Using `color_picker` in one of my projects, some users gave me feedback.
> They would find it useful if:
>  - We could input the hex value of the color directly 

I'd live this too for https://codeberg.org/OneTalker/OneTalker

PR #380, opened 2025-12-27
Testing update for color_picker

tonybierman:
The ColorPicker widget lacked `operate()` implementations, preventing iced_test from discovering and interacting with its buttons.  

## Remediation

1. Added `operate()` method to ColorPicker widget
2. Added `operate()` method to ColorPickerOverlay
3. Rewrote integration tests** (tests/color_picker_integration_tests.rs)

## What I Didn't Do

I didn't make the 6 color manipulation areas testable (HSV canvas, RGBA sliders). These are custom-drawn canvas elements, not widget components. The color sliders use custom canvas drawing and direct event handling, not the widget tree. Making them testable would require either:

- Refactoring to use Slider widgets (breaking change)
- Creating custom operation types for canvas interactions (complex)

It seems to me that the cost/benefit doesn't justify it - the critical user actions (opening picker, accepting/canceling) are now testable.

## Impact

- **No breaking changes** - purely additive internal implementations

### DIFF
--- a/src/widget/color_picker.rs
+++ b/src/widget/color_picker.rs
@@ -13,7 +13,10 @@ use iced_core::{
     layout::{Limits, Node},
     mouse::{self, Cursor},
     overlay, renderer,
-    widget::tree::{self, Tag, Tree},
+    widget::{
+        Operation,
+        tree::{self, Tag, Tree},
+    },
 };
 use iced_widget::Renderer;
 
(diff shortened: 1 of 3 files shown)


Ultraxime:
As you said, the internal working of the color picker is not really good. 
But as discussed in #377, the widget may be refactored (soon or not) to solve the issue of internal behaviours not being widgets.

genusistimelord (approved):
looks good thank you, Any other changes here can always come later by others if they want to make it better.

PR #396, opened 2026-01-01
Added code coverage configuration file

tonybierman:
### Summary

This is a small PR dealing with code coverage.  

- I've added `tarpaulin.toml` to configure code coverage runs, and updated the testing guidelines with a code coverage section.
- As of my most recent analysis, code coverage is sitting at about 44% lines covered.  As I state in the testing guide (`tests/README.md`), it is suggested that integration testing prioritize critical paths and common failure points rather than exhaustive coverage.  But metrics help us to know where we stand, so at any time detailed coverage reports can be seen in the `target/tarpaulin` folder after running:

```bash
cargo tarpaulin
```

### DIFF
--- a/tarpaulin.toml
+++ b/tarpaulin.toml
@@ -1,0 +1,6 @@
+[tarpaulin]
+out = ["Html", "Xml", "Markdown"]
+output-dir = "target/tarpaulin"
+exclude-files = ["examples/*", "tests/*"]
+skip-clean = true
+timeout = "5m"
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,6 +1,6 @@
-# Integration Tests Guide
+# Code Integration Testing Guide
 
-This directory contains integration tests for `iced_aw` widgets using the `iced_test` simulator framework.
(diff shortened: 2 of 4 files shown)


genusistimelord (approved):
Thank you. It is nice to see fuzzing code coverage and such but generally this is best done for unsafe or parsing code. With Graphical stuff Image Snapshots to detect regressions tends to be a better approach with bare minimal fuzzing just on the parts that make sense for it to Occur.  I heard a while back xilem used to do this to track regressions.  Anyways just wanted to let you know to try and reduce the work to only Really Require portions of the code that matter for these tests. Ones that would actually break in usage.   For example the Number input disallowing  above Max or Min fuzzing to ensure its working etc.

tonybierman:
> Thank you. It is nice to see fuzzing code coverage and such but generally this is best done for unsafe or parsing code. With Graphical stuff Image Snapshots to detect regressions tends to be a better approach with bare minimal fuzzing just on the parts that make sense for it to Occur. I heard a while back xilem used to do this to track regressions. Anyways just wanted to let you know to try and reduce the work to only Really Require portions of the code that matter for these tests. Ones that would actually break in usage. For example the Number input disallowing above Max or Min fuzzing to ensure its working etc.

It makes sense.

Issue #378, opened 2025-12-27
ContextMenu overlay does not show hover states


zquestz:
Buttons inside the ContextMenu overlay do not show hover states. Clicks work correctly.

**Reproduction:**

Run the `context_menu` example:

```bash
cargo run --example context_menu
```

Right-click to open the menu, then hover over the buttons. No hover effect is displayed. For comparison, the "right click me!" button outside the context menu shows hover correctly.

**Environment:**

- iced_aw version: 0.13.0
- iced version: 0.14.0
- OS: Arch

noamichael:
It seems the buttons in the context menu incorrectly have a `disabled` state.

I figured this out by adding a custom style for the buttons where I print out the `status` (`Pressed`, `Active`, `Hover`, or `Disabled`:

```rs
fn context_menu_button_style() -> impl Fn(&iced::Theme, button::Status) -> button::Style {
    move |theme: &iced::Theme, status: button::Status| {
        println!("Context menu button style called with status: {:?}", status);
        // other code...
    }
}
``` 

The only `status` the buttons will show are `disabled`:

```text
Context menu button style called with status: Disabled
```

This is strange as the status is only ever `disabled` if there is no `on_press` action:

https://github.com/iced-rs/iced/blob/595af03a9f1806f1052685081e50909d596dcdc8/widget/src/button.rs#L342

```rs
let current_status = if self.on_press.is_none() {
     Status::Disabled
}
```

genusistimelord:
im guessing we need to ensure they have an on_press event since iced has changed this.

noamichael:
The current `context_menu` example does have `on_press` so there must be something else going on:

https://github.com/iced-rs/iced_aw/blob/13467294b5289b017f23d6b014a8c84d7bbf70dd/examples/context_menu.rs#L64

I was also able to recreate this in my test app. I might poke around and see if I can spot why this would happen.

genusistimelord:
ok thank you for taking a look. its hard to tell what the issue is if it already has the on_press

noamichael:
I updated the example locally to only have a single button to make debugging easier. I'm probably a bit over my head for this one, but what I've discovered is the issue is related to how the status of the button is **unset** during the `draw` function, so it falls back to disabled:

<img width="1224" height="328" alt="Image" src="https://github.com/user-attachments/assets/4abf9dc1-926b-48fd-8099-da09ef3e4b4c" />

[...]

PR #399, opened 2026-01-02
Fix Issue #378

tonybierman:
### Background

Issue #378: Buttons inside the ContextMenu overlay do not show hover states. 

### Summary of Key Changes

Created a new MenuButton widget specifically designed for menus and overlays that:

1. Explicitly manages hover state - Tracks hover in its update() method by directly checking cursor.is_over(bounds)
2. Draws its own background - Renders hover effects directly in draw() instead of relying on iced's styling system

Files Modified/Created

- Created: src/menu_button.rs - The new widget
- Modified: src/widget.rs - Added module export
- Modified: src/lib.rs - Added public export
- Modified: Cargo.toml - Added menu_button feature
- Modified: examples/context_menu.rs - Updated to use MenuButton
- Modified: src/widget/overlay/context_menu.rs - Fixed warning

####  Usage
```rust
  use iced_aw::MenuButton;

  MenuButton::new(text("Menu Item"))
      .on_press(Message::ItemClicked)
      .width(Length::Fill)
      .padding(8.0)
```

3. Updated Example

- Modified examples/context_menu.rs to use MenuButton
(Right-click on "right click me!" and hover over the menu items - you should now see hover highlights.)

4. Updated integration tests
- Modified: tests/context_menu_integration_tests.rs
- Created: tests/menu_button_integration_tests.rs

```bash
# test
cargo test --lib widget::menu_button::tests && cargo test --test menu_button_integration_tests

# example 
cargo build --example context_menu --features menu_button,context_menu
```

### DIFF
--- a/src/widget/menu_button.rs
+++ b/src/widget/menu_button.rs
@@ -100,22 +100,18 @@ where
         Size::new(self.width, Length::Shrink)
     }
 
-    fn layout(
-        &mut self,
-        tree: &mut Tree,
-        renderer: &Renderer,
-        limits: &Limits,
-    ) -> Node {
+    fn layout(&mut self, tree: &mut Tree, renderer: &Renderer, limits: &Limits) -> Node {
         let padding = iced_core::Padding::from(self.padding);
         let limits = limits.width(self.width).shrink(padding);
 
-        let mut content = self.content.as_widget_mut().layout(
-            &mut tree.children[0],
-            renderer,
-            &limits,
-        );
+        let mut content =
+            self.content
+                .as_widget_mut()
+                .layout(&mut tree.children[0], renderer, &limits);
 
-        let size = limits.resolve(self.width, Length::Shrink, content.size()).expand(padding);
+        let size = limits
+            .resolve(self.width, Length::Shrink, content.size())
+            .expand(padding);
         content.move_to_mut(Point::new(self.padding, self.padding));
 
         Node::with_children(size, vec![content])
@@ -164,7 +160,10 @@ where
         );
 
         // Draw content
-        let content_layout = layout.children().next().expect("MenuButton should have content layout");
+        let content_layout = layout
+            .children()
+            .next()
+            .expect("MenuButton should have content layout");
         self.content.as_widget().draw(
             &tree.children[0],
             renderer,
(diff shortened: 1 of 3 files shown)


genusistimelord (approved):
thank you for the quick fix on this. I am guessing button couldn't be used?

tonybierman:
> thank you for the quick fix on this. I am guessing button couldn't be used?

I wanted to use Button, but I hit a lot of dead ends.  I have some opinions as to why it doesn't work, but the Iced team folks would probably know better than I could say.

I reviewed the MenuBar, and from what I can tell, it doesn't use Button either. It seems like MenuBar completely bypasses Iced's button styling system.  

Instead of relying on Iced to style buttons, MenuBar:
- Wraps arbitrary widgets (text, icons, etc.) in Item
- Tracks mouse position to determine which item is "active"
- Draws its own background quad directly using renderer.fill_quad()

So I took the same approach with the new MenuButton:
- Track hover state explicitly in State { is_hovered: bool }
- Draw background directly in draw() using renderer.fill_quad()
- Bypass Iced's button styling entirely

Why MenuButton is better for ContextMenu:
1. Simple mental model: "This is a button that works in overlays"
2. No external coordination: Each button is independent
3. Easy to use: Developers don't need to understand menu state tracking
4. Consistent API: Works like regular buttons, just with overlay support
5. Minimal code: Users just replace Button with MenuButton

zquestz:
This implementation is extremely buggy for me...

I am using a Context Menu in a table view, and it only draws correctly when inside the table view visible area, if its not over that view, it shows blank and does not draw correctly.

tonybierman:
@genusistimelord let’s back out the merge

genusistimelord:
> @genusistimelord let’s back out the merge

I reverted it back to the original. Probably need to ensure the newer widget can function in its own Overlay scope for rendering.

tonybierman:
> > @genusistimelord let’s back out the merge
> 
> I reverted it back to the original. Probably need to ensure the newer widget can function in its own Overlay scope for rendering.

Thanks.  Sorry for the trouble.

genusistimelord:
no trouble, this is just the grief of open source projects that are additions to another project.

genusistimelord:
@tonybierman  let me know if you need anything on this?

tonybierman:
@genusistimelord I haven't been working on this

PR #401, opened 2026-01-03
Snapshot hash match testing, CI workaround for Mac and Windows

tonybierman:
## Hash matching for snapshot tests
- Turned on hash matching for snapshot tests.  You'll see a baseline sha256 file next to each png in the tests/snapshots folder. 

## Workaround for snapshot tests in CI (Windows and Mac only)
- In the CI test workflow (test.yml), we run tests on Linux, Mac and Windows.  The assert_snapshot_matches() function has been passing on Linux, but failing on both Windows and Mac.  
- I confirmed it was not a path separator issue.  
- In my opinion, comparing raw RGBA pixel data will differ across platforms due to rendering differences.  Baselines taken on Linux are failing against tests taken on Windows and Mac.
- In the long run, one strategy might be to take baselines for all three - Linux, Windows and Mac.
- But for now, I've added logic to the `assert_snapshot_matches()` function that checks if the run is a CI workflow, and if so skip the test if Mac or Windows.  If CI Linux, tests run normal.  If local run (not CI), tests run normal on all OS.

```bash
cargo test --tests --bins --lib
```

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -260,33 +260,41 @@ macro_rules! test_helpers {
             ui: &mut iced_test::Simulator<'_, $message_type>,
             baseline_name: &str,
         ) -> Result<(), iced_test::Error> {
-            // Skip snapshot matching on non-Linux platforms
-            #[cfg(not(target_os = "linux"))]
+            // Skip snapshot matching on macOS and Windows in CI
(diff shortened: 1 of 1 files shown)


genusistimelord (approved):
looks good and yeah sucks that it renders differently on each OS. Probably should not be this way but sadly it is.

PR #418, opened 2026-03-28
Scope `menu_items` macro

bicarlsen:
Scopes `menu_items` macro in `menu_bar` so users don't need to manually import it.


genusistimelord (approved):
thank you

PR #420, opened 2026-04-26
Add safe triangle behavior for nested menus

zao111222333:
I inroduced the safe triangle behavior for menu interactions. It tracks the last cursor position  over a parent menu item, builds a triangular safe area toward the opened submenu, and keeps the  submenu open while the cursor moves through that area. This prevents nested menus from closing  prematurely when users move the cursor from a parent item into its child menu. 

Demonstrated by 
```
cargo run --release --example menu
```

**Before**:

https://github.com/user-attachments/assets/b4ee8130-6117-4da6-94ba-bb848be8d586

**Now**:

https://github.com/user-attachments/assets/42a855ae-c61d-4b86-a8be-9f9cd104c989

### DIFF
--- a/src/widget/menu/common.rs
+++ b/src/widget/menu/common.rs
@@ -319,3 +319,53 @@ macro_rules! itl_iter_slice_enum {
     };
 }
 pub(super) use itl_iter_slice_enum;
+
+#[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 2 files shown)


genusistimelord (approved):
thank you this looks good.

PR #440, opened 2026-08-01
Deps(iced)/text input change

tsuza:
I've updated `iced_aw` to the latest `Iced` changes ( namely this PR https://github.com/iced-rs/iced/pull/3404, more specifically https://github.com/iced-rs/iced/tree/4cad51ba61e3d76a7e696ad5923d675a20994e34 ). I have not committed the `Cargo.lock` file.

I've rewritten `NumberInput` from scratch thanks to the changes from the PR. It behaves and looks the same way, and this time around copying and pasting works. It also feels much faster compared to the previous one. I noticed, at least in debugging mode, that it'd heavily lag whenever you tried to delete something. Some things to note:
- The default behaviour of `TextInput` is to `Fill` the available width. Since `NumberInput` is just a specialized `TextInput` I've kept this behaviour. If needed, we can change it to match the old behaviour.
- I've slightly changed the API ( mainly `NumberInput::new` )  to match `TextInput`. Now it also accepts a placeholder, and you need to call the methods to set the bounds and `on_input`. I think it makes more sense, but it's of course a breaking change.
- I've removed `ignore_scroll`.
- I've removed a couple of tests since they didn't make sense or work anymore. Most of them have been kept and pass.
- `TypedInput` has been completely removed.

Whilst I'm here, I'm proposing another breaking change in the API: both this rewrite and the previous version currently have an issue when deleting a value from the input field. If you have `5` and delete it, the field will appear empty visually, but the value in the app state will still remain `5`. This becomes especially confusing when you try to paste something ( without updating the value in your app ) and see the previous value reappear in the field. `on_input` should return an `Option<T>` in these cases where the input field is emptied, since, unlike strings, there is no "empty" number.


Ultraxime:
Hi, 

I know that I created `TypedInput`  as a workaround originally, but I grew to like it and I know often use it in my gui, so I don't see any reason to remove it. 

As you stated there is no empty numbers, that why it was for me an expected behaviour to always have a value especially the last one when a value inside the cell is invalid, including when there is nothing. 
The field is a number and should always be a number.

tsuza:
The developer won't be able to tell if a field has been completely emptied with the current behaviour, on top of having a desync between the value in the app state and what is being shown, which I think goes against Iced's philosophy. I feel like we can hit two birds with one stone by changing the return type to `Option<T>`. By doing it, the developer can keep the same behaviour as before if needed.

Putting the workaround for `NumberInput` aside, does `TypedInput` do something more compared to `TextInput`? If not, I feel like I'd be reasonable to remove it so we also have less widgets to maintain.

Nevertheless, thanks for the insight.

Ultraxime:
For me the status of the `NumberInput` is a number and a string . 
Moreover the idea is to allow currently misspelled number. For example `1.` is not a valid float or is equal to `1` I don't remember, so it could cause issues, either being misread for and empty field or just `1` preventing from writing any digit after the dot. 
The idea we had was that on submit, the dev could check if it's empty or anything. 
But a `on_maybe_input` could be nice too, to have your example. 

I don't know if many people use the `TypedInput` but I do, and I rather maintain it here than in all my project. 
Concerning it's design it's mostly a `TextInput` but that return the value and not the text. 
It was thought to change it's theme when the value is invalid but I didn't master enough theming at the time to do it.

tsuza:
So it's a more generic `TextInput` or `NumberInput`, nothing else? In that case, I guess I could port it over if needed.

tsuza:
I'll need to take a closer look to the buttons.
image

tsuza:
Should be an improvement:
image

Here is how it looks with the default styling in the example:
image

tsuza:
~On a side note, can we update `Cargo.lock` to avoid having to pull, for example, `iced_renderer v0.14.0`, since we're already targeting master anyway?~

Scratch that, `iced_fonts` was the culprit.

Redhawk18:
> ~On a side note, can we update `Cargo.lock` to avoid having to pull, for example, `iced_renderer v0.14.0`, since we're already targeting master anyway?~
> 
> Scratch that, `iced_fonts` was the culprit.

Oops

Redhawk18:
One of these does need to go. @tsuza @Ultraxime 

It's silly to have a filtered input and a number filtered input

Ultraxime:
As I previously stated for me typed and number input each provid a different service . 
TypedInput is designed to create a filtered input on a FromStr object, ideally changing the theme when the value is not valid . 
On the other hand NumberInput is a specialised version of the previous, supporting only objects which behave like numbers (can be increased and decreased) and doesn't allow misformed value (except special case)

tsuza:
With the introduction of [components](https://github.com/iced-rs/iced/pull/3435), I feel like we could (re)rework this altogether. At the end of the day the logic of the widget itself is mostly the same ( it's just a text input ), so thanks to the component system we can just handle the necessary logic in `update` and add the increase & decrease buttons with a `stack`.

This would greatly simplify the widgets, and we could look into implementing both. I'd gladly do it, but right now I'm kind of busy so I'd rather put this off for a couple of weeks if it's not an issue.

Ultraxime:
It sounds like a valid approach to me. 

I don't think we are in a rush and I find myself too busy too to work on it.

Redhawk18:
> As I previously stated for me typed and number input each provid a different service . TypedInput is designed to create a filtered input on a FromStr object, ideally changing the theme when the value is not valid . On the other hand NumberInput is a specialised version of the previous, supporting only objects which behave like numbers (can be increased and decreased) and doesn't allow misformed value (except special case)

This can be generic-fied, I stand by what I said before. Number input could just be a version of the typedinput offered predefined. Then users can impl a increase/decrease trait if they want the arrows on the typed input, otherwise those arrows are not drawn.

tsuza:
Sounds good to me, then.

Ultraxime:
> > As I previously stated for me typed and number input each provid a different service . TypedInput is designed to create a filtered input on a FromStr object, ideally changing the theme when the value is not valid . On the other hand NumberInput is a specialised version of the previous, supporting only objects which behave like numbers (can be increased and decreased) and doesn't allow misformed value (except special case)
> 
> This can be generic-fied, I stand by what I said before. Number input could just be a version of the typedinput offered predefined. Then users can impl a increase/decrease trait if they want the arrows on the typed input, otherwise those arrows are not drawn.

I agree with you 
And the behaviour of allowing invalid input could be a flag on the widget

PR #442, opened 2026-08-06
selection_list: accept Borrow<[T]> instead of &[T] for options

### LINKED ISSUE #411: Make SelectionList take a Borrow<[T]> instead
`PickList` bundled with iced takes the options as a `Borrow<[T]>` which also accepts owned types. 

This is very convenient as otherwise you must store the options with the same lifetime as the app state, and cannot produce these values dynamically inside of a `view` method.

Joshuaisikah:
## Summary

Closes #411.

`SelectionList` currently takes `options: &'a [T]`, which forces callers to store the options with the same lifetime as their app state — unlike `PickList`, which already takes `L: Borrow<[T]>` and happily accepts an owned `Vec<T>` built inline in `view()`.

This mirrors that same pattern for `SelectionList`:
- `SelectionList<'a, T, L, Message, Theme, Renderer>` is now generic over `L: Borrow<[T]>`.
- The previous implementation eagerly built and cached a `Container<Scrollable<List>>` once at construction time, embedding the borrowed `&'a [T]` into it directly — exactly the constraint the issue wants removed. Since `options` is now an owned `L`, that can no longer be pre-built and cached (the inner `List` widget needs a live `&[T]`, not an owned `L`).
- Restructured so the container is rebuilt on demand from a private `list_container()` helper, called from each `Widget` trait method (`diff`/`layout`/`update`/`mouse_interaction`/`draw`/`operate`) — the same pattern `PickList` already uses for its menu overlay.
- The internal `List` widget now takes `class` and `on_selected` by reference rather than by value, since they're borrowed from `SelectionList`'s own fields on each call instead of being moved in once at construction.
- `widget::helpers::selection_list[_with]` updated to thread the new `L` parameter through.
- Added a regression test (`selection_list_accepts_owned_vec`) that builds an owned `Vec<String>` inline and constructs a `SelectionList` from it directly, proving the actual use case from the issue.

## Test plan
- [x] `cargo test --features selection_list --lib selection_list` — all 25 unit tests pass, including the new owned-`Vec` regression test
- [x] `cargo clippy --all-features --lib --tests --bins` — no new warnings (same 5 pre-existing warnings in unrelated widgets: card, tabs, sidebar, menu_bar)
- [x] `cargo test --all --all-features --lib --tests --bins` — all `selection_list` integration tests pass except `selection_list_can_find_all_options_in_list`, a snapshot-diff test; confirmed this is **pre-existing environment flakiness** by reproducing the identical failure against the unmodified `main` baseline, and by observing 10 other unrelated widgets' snapshot tests (badge, card, color_picker, date_picker, number_input, sidebar, tab_bar, time_picker, wrap, etc.) fail the same way in this environment
- [x] `cargo build --example selection_list --features selection_list` — existing example builds unmodified, confirming the common `&[T]` usage still infers `L` correctly with no call-site changes needed
- [x] `cargo fmt --check` — clean

### DIFF
--- a/src/widget/helpers.rs
+++ b/src/widget/helpers.rs
@@ -333,15 +333,15 @@ where
 ///
 /// [`SelectionList`]: crate::SelectionList
 #[must_use]
-pub fn selection_list_with<'a, T, Message, Theme, Renderer>(
-    options: &'a [T],
+pub fn selection_list_with<'a, T, L, Message, Theme, Renderer>(
+    options: L,
     on_selected: impl Fn(usize, T) -> Message + 'static,
     text_size: f32,
     padding: impl Into<Padding>,
-    style: impl Fn(&Theme, Status) -> crate::style::selection_list::Style + 'a + Clone,
+    style: impl Fn(&Theme, Status) -> crate::style::selection_list::Style + 'a,
     selected: Option<usize>,
     font: iced_core::Font,
-) -> crate::SelectionList<'a, T, Message, Theme, Renderer>
+) -> crate::SelectionList<'a, T, L, Message, Theme, Renderer>
 where
     Message: 'a + Clone,
     Renderer: 'a + renderer::Renderer + iced_core::text::Renderer<Font = iced_core::Font>,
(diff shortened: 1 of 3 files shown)


Redhawk18:
Thanks claude

